PR #64, opened 2020-04-19
WIP: Remove crop from HRF example

rob-luke:


rob-luke:
Fails with `MemoryError: Unable to allocate 4.02 GiB for an array with shape (23234, 23239) and data type float64
307
`

larsoner:
Where does the memory error happen? `nilearn`?

I wonder if they need to use a sparse representation internally to avoid this...

rob-luke:
Haven't dug in to this yet. My initial hope that switching from Nistats to Nilearn would magically fix the issue, but I think the code base is largely the same.I plan to dig in to this when I get a chance, but its currently second priority to ensuring my code is correct. So I have raised this PR to remind myself and/or demonstrate to anyone interested where the issue is.

rob-luke:
Closing in favour of #191

Issue #101, opened 2020-06-01
Region of interest analysis


rob-luke:
Currently statistics are computed per channel. Need to implement a way to collapse multiple channels to a region of interest analysis.

The computation/statistics are described in equations 36-37 in section 6.3 of [Santosa, Hendrik, et al. "The NIRS brain AnalyzIR toolbox." Algorithms 11.5 (2018): 73](https://www.mdpi.com/1999-4893/11/5/73/htm).

rob-luke:
Regions of interest should be specified in the same manner as in [MNE plot image](https://mne.tools/dev/generated/mne.BaseEpochs.html?highlight=plot_image#mne.BaseEpochs.plot_image)...

```python
group_by=dict(Left_ROI=[1, 2, 3, 4], Right_ROI=[5, 6, 7, 8])
```

rob-luke:
Closed in #137

PR #111, opened 2020-06-13
WIP: How should I store regression results?

rob-luke:
Running a GLM on our NIRS data using Nilearn returns a `nilearn.stats.regression.RegressionResults` and `labels` array. In this PR I am trying to determine how I should store this returned results in a MNE style way.

Previously I have been returning the two results directly to the user, then I provided a convenience function (`mne_nirs.utils._GLM_to_tidy_long(raw, labels, RegressionResults)`) to wrangle the data and return a tidy data frame (see example usage [here](https://mne.tools/mne-nirs/auto_examples/plot_10_hrf.html#export-results)).

This approach seems messy as you have to pass around three things (`mne.Raw`, `labels`, `RegressionResults`) and the returned regression results have a slightly tricky nested structure, which I would prefer to hide from the end user. So I would like to determine if there is a way I can store these results in the MNE data structure.

However, I would prefer not to write my own stats storage structure as I believe the `nilearn.stats.regression.RegressionResults` format is well considered, well implemented and documented. To start a discussion I have committed an example approach where I put the regression results in to a list then store this in `raw._data`, I am sure this is not the best way.

@larsoner and @agramfort do you have any suggestions on how I should store GLM results returned from Nilearn in MNE?

<details>
  <summary>Detailed issue (and error) with current PR approach</summary>
  
If I use the approach as committed in this PR then I get the following error if I try to call the returned modified raw object.

```python

---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
~/Library/Python/3.7/lib/python/site-packages/IPython/core/formatters.py in __call__(self, obj)
    700                 type_pprinters=self.type_printers,
    701                 deferred_pprinters=self.deferred_printers)
--> 702             printer.pretty(obj)
    703             printer.flush()
    704             return stream.getvalue()

~/Library/Python/3.7/lib/python/site-packages/IPython/lib/pretty.py in pretty(self, obj)
    392                         if cls is not object \
    393                                 and callable(cls.__dict__.get('__repr__')):
--> 394                             return _repr_pprint(obj, self, cycle)
    395 
    396             return _default_pprint(obj, self, cycle)

~/Library/Python/3.7/lib/python/site-packages/IPython/lib/pretty.py in _repr_pprint(obj, p, cycle)
    682     """A pprint that just redirects to the normal repr function."""
    683     # Find newlines and replace them with p.break_()
--> 684     output = repr(obj)
    685     lines = output.splitlines()
    686     with p.group():

~/Documents/Repositories/mne-python/mne/io/base.py in __repr__(self)
   1628         name = self.filenames[0]
   1629         name = '' if name is None else op.basename(name) + ', '
-> 1630         size_str = str(sizeof_fmt(self._size))  # str in case it fails -> None
   1631         size_str += ', data%s loaded' % ('' if self.preload else ' not')
   1632         s = ('%s%s x %s (%0.1f s), ~%s'

~/Documents/Repositories/mne-python/mne/utils/mixin.py in _size(self)
     50             size += object_size(self.data)
     51         elif hasattr(self, '_data'):
---> 52             size += object_size(self._data)
     53         return size
     54 

~/Documents/Repositories/mne-python/mne/utils/numerics.py in object_size(x)
    688             size += object_size(value)
    689     elif isinstance(x, (list, tuple)):
--> 690         size = sys.getsizeof(x) + sum(object_size(xx) for xx in x)
    691     elif isinstance(x, datetime):
    692         size = object_size(_dt_to_stamp(x))

~/Documents/Repositories/mne-python/mne/utils/numerics.py in <genexpr>(.0)
    688             size += object_size(value)
    689     elif isinstance(x, (list, tuple)):
--> 690         size = sys.getsizeof(x) + sum(object_size(xx) for xx in x)
    691     elif isinstance(x, datetime):
    692         size = object_size(_dt_to_stamp(x))

~/Documents/Repositories/mne-python/mne/utils/numerics.py in object_size(x)
    695                    for xx in [x, x.data, x.indices, x.indptr])
    696     else:
--> 697         raise RuntimeError('unsupported type: %s (%s)' % (type(x), x))
    698     return size
    699 

RuntimeError: unsupported type: <class 'nilearn.stats.regression.RegressionResults'> (<nilearn.stats.regression.RegressionResults object at 0x154b997b8>)

```

</details>


larsoner:
>  To start a discussion I have committed an example approach where I put the regression results in to a list then store this in raw._data, I am sure this is not the best way.

Indeed `Raw._data` should only ever be `np.ndarray` so this will not work.

Have you looked at the [linear regression](https://mne.tools/dev/generated/mne.stats.linear_regression.html#mne.stats.linear_regression) examples in MNE? Any ideas that come from looking at those use cases?

If this API will not work, one thing I notice is that the nilearn class has a [simple constructor](https://github.com/nilearn/nilearn/blob/46c7d2102eccad9d9dad0a975741ee3bfdbed87a/nilearn/stats/regression.py#L286-L308), so we can probably fairly easily [save it to disk using h5io](https://github.com/mne-tools/mne-python/blob/master/mne/morph.py#L526-L529) and then [load and reconstruct it](https://github.com/mne-tools/mne-python/blob/master/mne/morph.py#L591-L601) like we do for the `sdr_morph` and `pre_affine` classes of the SourceMorph object, which are actually `nipy` morphing objects.

Would it make sense to add the `labels` as an attribute at the nilearn end? If they're open to this, then it seems like the simplest solution is to have `labels` become a part of that class, we (or they) provide a simple I/O wrapper based on HDF5, then you just need to keep your `raw` and your `RegressionResults` around (separately and combine when appropriate).

rob-luke:
> Have you looked at the linear regression examples in MNE? Any ideas that come from looking at those use cases?

Thanks @larsoner I had missed this. I considered MNE linear regression example output and expanding the dictionary to include covariance matrices (required for creating regions of interest #101). However, I could not see an inbuilt way in MNE to do contrast statistics with this output. Whereas Nilearn `RegressionResults` gives us this for free (see [docs](https://nistats.github.io/modules/generated/nistats.regression.RegressionResults.html#nistats.regression.RegressionResults.Fcontrast)). So I am thinking to stick with RegressionResults so I don't have to program these stats myself.

I have modified my approach and am now creating my own dictionary by passing in each channel one by one, then I can name the dict entries to match the MNE channel names. This avoids the Nilearn nesting and allows me to easily address the regression results (to ensure I am matching channels correctly with the raw.info) without passing around the labels parameter.

Do you foresee any large issues with this approach? I think I am happy with it now (still code and doc tweaks required).

mne_nirs/statistics/_run_GLM.py
```diff
@@ -28,16 +31,28 @@ def run_GLM(raw, design_matrix, noise_model='ar1', bins=100,
...
+
+    return results
+
 
-    return labels, glm_estimates
+def compute_contrast(glm_est, contrast):
+    from nilearn.stats.contrasts import compute_contrast as _cc
+    a = np.array(list(glm_est.keys()))
+    cont_result = _cc(a, glm_est, contrast)
+    return cont_result
```
rob-luke, line 58: If we use the Nilearn structure then we get these nice contrast computations for free. It only requires a small wrangling function.

rob-luke: Example in docs can be found here...
https://247-254026779-gh.circle-artifacts.com/0/dev/auto_examples/plot_10_hrf.html#compute-contrasts

which produces the expected contralateral activation topo map

image

rob-luke:
> one thing I notice is that the nilearn class has a simple constructor, so we ....

Ok I think I understand this now. Sounds good and seems the `nilearn.stats.regression.RegressionResults` fits this approach.

@larsoner I am quite happy with the current approach of putting the nilearn regression results in a dict and passing that around with raw. But I would prefer to check if you foresee any large issues with this approach before I go any further down this coding rabbit hole. Thanks

larsoner:
> I am quite happy with the current approach of putting the nilearn regression results in a dict and passing that around with raw. But I would prefer to check if you foresee any large issues with this approach before I go any further down this coding rabbit hole. Thanks

It depends a bit on what you mean by "passing it around with raw". But as long as there is a good method for saving the `nilearn` class *with the labels* (would require some work / convincing at their end), then telling people to save that and the `raw` independently seems fine.

Do you want to update a MNE-nirs example to show how you envision users using this API / saving to get back to an analysis / coefficients that they previously computed?

rob-luke:
> Do you want to update a MNE-nirs example to show how you envision users using this API / saving to get back to an analysis / coefficients that they previously computed?

I've never done such a thing myself. I've always restarted computation from scratch as even my largest group studies take less than 5 min to compute. But this is something i will look in to.

Sounds like no large objections here, so I will push forward and be sure to investigate how to save analysis so users can resume later.

doc/api.rst
```diff
@@ -50,6 +50,7 @@ Statistics
    :toctree: generated/
 
    run_GLM
+   compute_contrast
```
larsoner, line 53: You can add `:orphan:` at the top of the file to get the "document isn't included in any toctree" comment to go away

mne_nirs/utils/_io.py
```diff
@@ -3,25 +3,91 @@
...
+    Returns
+    -------
+    df : Tidy data frame,
+        Data from statistic object in tidy data form.
+    """
+
+    if isinstance(statistic, dict):
+        if isinstance(statistic[list(statistic.keys())[0]],
+                      nilearn.stats.regression.RegressionResults):
+            df = _tidy_RegressionResults(raw, statistic, design_matrix)
```
larsoner, line 36: If the first `if` evals to True and the second to False, `df` will not be in the namespace when `return df` is called. I would make this a `if a and b` clause instead

mne_nirs/utils/_io.py
```diff
@@ -3,25 +3,91 @@
...
+
+    elif isinstance(statistic, nilearn.stats.contrasts.Contrast):
+        df = _tidy_Contrast(raw, statistic, design_matrix)
+
+    else:
+        warn("Unknown GLM result type")
```
larsoner, line 42: If you hit this code path, df will not be defined when `return df` is called

I would raise an error here instead

larsoner:
> I've never done such a thing myself. I've always restarted computation from scratch as even my largest group studies take less than 5 min to compute. But this is something i will look in to.

No need here, can always add it later. Just want to consider these things during design so that later I/O implementation is easier

doc/api.rst
rob-luke: I've been wondering that for ages!!!

Issue #113, opened 2020-06-14
Make mini gallery appear under each function documentation on website


rob-luke:
In the API section of the website, when the page for each function is open I would like it to have a list of examples that use the function.

I looked here but couldn't get it working... https://sphinx-gallery.github.io/stable/configuration.html#add-mini-galleries-for-api-documentation

agramfort:
copy this files from mne python

_templates/class.rst
_templates/function.rst

maybe this will work. If not @larsoner may spot the issue quickly

>

rob-luke:
Thanks @agramfort. I have opened #114 to build this in the ci and incase it helps highlight where my issue is.

PR #114, opened 2020-06-14
WIP: Make mini gallery appear under each function documentation on website

rob-luke:
As described in #113


rob-luke:
Still doesn't work :(

larsoner:
Your templates are in ` doc/_templates/class.rst`, which means that they will *not* be automatically detected by `autosummary`. You either need to pass `:template:` every time you want to autodoc a class (pain) or move these to `doc/_templates/autosummary/class.rst` / `function.rst` as we do in MNE-Python (automagical, easier)

rob-luke:
Woohoo it works! Thanks @larsoner and @agramfort

PR #115, opened 2020-06-14
Update environment and requirements to match MNE

rob-luke:


rob-luke:
Fails on linux with 

```bash
E ImportError: libGLX.so.0: cannot open shared object file: No such file or directory
```

larsoner:
The advantage of using a config like MNE-Python is that you can benefit from our fixes, `meld` or do a similar diff whenever you see problems like this and hopefully this would pop up as a candidate:

https://github.com/mne-tools/mne-python/blame/master/.circleci/config.yml#L58

In this case it's probably `libglx0`. FYI this is really a VTK bug that they're in the process of fixing but did not make it into the 9.0.0 wheels

larsoner:
Oh sorry this is Travis not CircleCI, but the fix is similar

https://github.com/mne-tools/mne-python/blame/master/.travis.yml#L12

larsoner:
(Or you could subscribe to all MNE-Python updates, but then you will have the same painful email load as @agramfort :) )

rob-luke:
> Or you could subscribe to all MNE-Python updates

Haha i do. Thats how I saw that MNE had updated pyvista and thats what inspired this change here.

> The advantage of using a config like MNE-Python is that you can benefit from our fixes, meld 

Thanks, I need to get my tooling better. I should have some assisted way of tracking changes in relevant MNE files so I don't need to do this manually. Maybe once the API settles and this is totally usable Ill then get around to it.

> Oh sorry this is Travis not CircleCI, but the fix is similar

Thank you! Appreciate the pointer

Issue #117, opened 2020-06-17
Add a maximum distance parameter to get_long_channels


rob-luke:
Having a source-detector separation that is too large is detrimental to signal quality. `get_long_channels` should be modified to take a maximum distance parameter with default value of 35 mm.

PR #121, opened 2020-06-18
Added upper limit to long channel extraction

mshader:
Addresses #117

### DIFF
--- a/mne_nirs/channels/_short.py
+++ b/mne_nirs/channels/_short.py
@@ -38,7 +38,7 @@ def get_short_channels(raw, max_dist=0.01):
     return short_chans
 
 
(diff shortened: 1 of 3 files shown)


mne_nirs/channels/_short.py
```diff
@@ -65,5 +67,6 @@ def get_long_channels(raw, min_dist=0.01):
 
     dists = source_detector_distances(long_chans.info, picks=picks)
     long_chans.pick(picks[dists > min_dist])
+    long_chans.pick(picks[dists < max_dist])
```
rob-luke, line 70: This isn’t a testing issue, @mshader & @nattynoodles there is actually a bug here. Let’s go through this in the lab next week.

PR #146, opened 2020-07-28
WIP: Add snirf writer

rob-luke:
Add support for writing SNIRF files.

https://github.com/mne-tools/mne-python/issues/8060

PR #151, opened 2020-07-30
WIP: Add snirf writer (moved)

rob-luke:
Moved from #146 to my fork.

Add support for writing SNIRF files.

mne-tools/mne-python#8060


rob-luke:
These tests pass locally with the branch from https://github.com/mne-tools/mne-python/pull/8079

rob-luke:
@agramfort and @larsoner could you please review again. I would appreciate feedback if you think the method of saving `raw.info['dig']` is appropriate. Could you please review in parallel with https://github.com/mne-tools/mne-python/pull/8079

mne_nirs/io/snirf/tests/test_snirf.py
```diff
@@ -0,0 +1,53 @@
...
+
+    # Check info object is the same
+    obj_diff = object_diff(raw.info, raw_orig.info)
+    num_diffs = 0
+    for line in obj_diff.splitlines():
+        if ('logno' not in line) and ('scanno' not in line):
+            # logno and scanno are not used in processing
+            num_diffs += 1
+            print(line)
+    assert num_diffs == 0
```
larsoner, line 53: I would do this instead so it shows up directly in the error:
```suggestion
    diffs = ''
    for line in obj_diff.splitlines():
        if ('logno' not in line) and ('scanno' not in line):
            # logno and scanno are not used in processing
            diffs += f'\n{line}'
    assert diffs == ''
```
But I'm a bit surprised that the scanno and logno actually change. Seems like those should stay the same

mne_nirs/io/snirf/_snirf.py
```diff
@@ -0,0 +1,137 @@
+# Authors: Robert Luke <mail@robertluke.net>
+#
+# License: BSD (3-clause)
+
+import h5py as h5py
+from re import match as mch
```
agramfort, line 6: I would not use this custom shortcut. It required me to search in the file to know what `mch` was. I would use re.match to be explicit. It's not is too many places anyway

PR #159, opened 2020-08-27
Why is CI failing to build docs?

rob-luke:
Its building locally.

Its failing on this file, but is that the reason?


larsoner:
Warnings are treated as errors, and there is this:
```
generating gallery for auto_examples... [ 60%] plot_10_hrf.py

Extension error:
Handler <function generate_gallery_rst at 0x7fe6548b0e50> for event 'builder-inited' threw an exception (exception: zero-size array to reduction operation minimum which has no identity)
```
Though really this shouldn't happen -- `sphinx-gallery` should catch the error. So I wonder if it's an error within sphinx-gallery itself.

I'm going to restart the https://app.circleci.com/pipelines/github/mne-tools/mne-nirs/441/workflows/dc50c665-0471-4305-9043-d0f7f2e0d7e0/jobs/448 job with SSH and try running it myself to see if I can figure it out

larsoner:
The real error (using `make html SPHINXOPTS=-vv` after running `source $BASH_ENV` in circle SSH session) turns out to be a zero-size reduction error in `fig.savefig` in the matplotlib scraper. I'll push a commit to revert the naming and use https://github.com/sphinx-gallery/sphinx-gallery/pull/737, and from there hopefully you can figure out what code block / plotting is causing the problem. Usually it means that some figure or axis or something has zero size.

doc/requirements.txt
```diff
@@ -1,4 +1,4 @@
-sphinx_gallery
+https://github.com/larsoner/sphinx-gallery/archive/scraper_broken.zip
```
larsoner, line 1: @rob-luke this file seems to be redundant with the root `requirements_doc.txt`. This is not DRY and made it more challenging to debug this issue. It would be good to deduplicate. Personally I would just `cp requirements_doc.txt doc/requirements.txt`

larsoner:
@rob-luke you might also consider trimming `environment.yml` and `requirements.txt` to just what's needed for MNE-NIRS. For example it looks like a lot of build time is spent in building `statsmodels`. I'll adjust CircleCI to use `requirements_doc.txt` so it is closer at least to the speed of the GitHub action.

larsoner:
Here is your failure:

https://459-254026779-gh.circle-artifacts.com/0/dev/auto_examples/plot_10_hrf.html#analyse-regions-of-interest

Locally on macOS I get this output (at least when working around https://github.com/nilearn/nilearn/issues/2560) at that failure point:

Figure_1

So not sure how to fix it, but at least now you have a target to look into @rob-luke ! For example you could try SSH'ing into the Circle build and running the example yourself after `source $BASH_ENV; set +e` or so and then adding a `plt.savefig('temp.png')` on the `sns`-produced figure run with `MPLBACKEND=agg python -i examples/plot_10_hrf.py` and then `import pdb; pdb.pm()`'ing after it hits ...

rob-luke:
Thanks @larsoner! I really appreciate you looking at this, I was banging my head against the wall. You have raised a few issues which I will address...

> Locally on macOS I get this output

Ok this is a relief! That is exactly what's expected and shown in the last successfully built docs (a few weeks back) https://mne.tools/mne-nirs/auto_examples/plot_10_hrf.html#analyse-regions-of-interest

> So not sure how to fix it, but at least now you have a target to look into

Thank you very much!

> you might also consider trimming environment.yml and requirements.txt to just what's needed for MNE-NIRS. 

Yes agreed. I have been simply copying the MNE versions and adding nilearn master. This way I was getting the benefit of the teams work over in mne-python. But I would like to make this simpler as you suggest. I will raise an issue for this. I wonder if there is a way to load another requirements file, so I could point to the mne master requirements, then just add the additional things I need.

Generally I would like to get this process type work tidied before making a first release. I was planning to give this package a big push after the next MNE release and lock it down to 0.21. Then when nilearn cut their next release I would like to release v0.0.1 of this package.

Issue #160, opened 2020-08-28
Simplify requirements and environment.yml


rob-luke:
As discussed in #159 

> you might also consider trimming environment.yml and requirements.txt to just what's needed for MNE-NIRS.

rob-luke:
Can I make the tests run weekly so that we catch upstream bugs earlier, rather than noticing when someone opens a PR?

larsoner:
Yes, most CIs provide a way to run on a weekly/daily/whatever schedule in addition to being triggered by PRs

rob-luke:
Thanks I'll check this out. Excuse the notes to myself here, it helps me not loose track of issues.

PR #161, opened 2020-08-28
Can we just use MNE in requirements?

rob-luke:
Issue #160


rob-luke:
Need to rebase once paths are fixed.

rob-luke:
@larsoner do you foresee any issues with me going this sparse with the requirements and environment files?

agramfort:
installing mne-python like this will only give you numpy, scipy and matplotlib. You will not have any 3D support.

rob-luke:
Thanks @agramfort that is very relevant for me to know because MNE-nirs provides some nirs specific 3D plots.

It seems I shouldn't go bare minimum then. I will add the 3D libraries back in.

rob-luke:
Thanks for point this out @agramfort, I know see that in the CI tests the 3d plots are being skipped due to the `@requires_pysurfer` line.

image

@larsoner I am now struggling to see the middle ground here. I find the most helpful part of having the environment.yml file is that new users load that in to anaconda and they are instantly ready to go, which makes me think keeping the full file will make my life easier. Also this makes it easy for me to keep in sync with MNE. But as you mentioned, this is probably overkill for CI testing. Is it ok to leave as is for now and revisit this later?

larsoner:
Yeah it should be fine to leave it

rob-luke:
Thanks @larsoner. I would like to optimise these still (maybe when I've got a bit more time) so I will leave the issue open.

PR #162, opened 2020-08-31
Update plot_11_hrf_simulation.py

NathalieGou:
added a note for clarity to the 'simulate noisy NIRS data (white)' section


rob-luke:
Thanks @NathalieGou

PR #164, opened 2020-09-01
Point back to sphinx gallery master

rob-luke:


rob-luke:
Thanks @larsoner for the sphinx gallery fix and approval here.

PR #165, opened 2020-09-03
Enable GitHub codespaces

rob-luke:
Works for me. This would allow people to fork, then hit open in Codespaces, and be editing within a minute without installing anything locally.

Running pytest in the Codespaces vscode instance had all tests pass.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,27 @@
+FROM continuumio/miniconda3
+
+# Options for common setup script
+ARG INSTALL_ZSH="true"
+ARG UPGRADE_PACKAGES="false"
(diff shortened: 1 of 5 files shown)


larsoner:
@rob-luke once you have this working, would it be worth adding to MNE-Python, too? Not familiar with this functionality. Also not sure how it relates to binder:

https://sphinx-gallery.github.io/stable/configuration.html#binder-links

rob-luke:
@larsoner this is actually super handy. When viewing the rep you hit the big green code button (same as to clone) and choose launch workspace. It spins up a docker image configured already to have MNE and launches a web vs code front end. It has a nice terminal and you can code, run tests etc. I will test again then open PR upstream

Issue #77, opened 2020-04-27
Add stats second level model


rob-luke:
* https://github.com/nilearn/nilearn/issues/2372
* https://github.com/nilearn/nilearn/blob/master/nilearn/stats/second_level_model/second_level_model.py
* https://www.statsmodels.org/stable/generated/statsmodels.genmod.generalized_linear_model.GLM.html

Require:
* Dataset
* Add tutorial

rob-luke:
This is a first attempt

```python
from statsmodels.formula.api import rlm
rlm_model = rlm('theta ~ -1  + condition:Chroma:ch_name', input, groups=input["ID"]).fit()
print(rlm_model.summary())
```

rob-luke:
Only the snipper from above is required. So perhaps a simple example will be sufficient, and do not add a function.

rob-luke:
Closed in #167

PR #167, opened 2020-09-21
MRG: Add group analysis example

rob-luke:
Addresses #77 

View latest render of example at... 

https://576-254026779-gh.circle-artifacts.com/0/dev/auto_examples/plot_12_group_glm.html

To do:

- [x] Keep everything in SI units until plotting.
- [x] Tests.
- [x] Improve text.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -72,11 +72,18 @@ jobs:
 
         # Build docs
         - run:
+            name: Download data
+            command: |
+              cd examples;
+              git clone --depth 1 https://github.com/rob-luke/BIDS-NIRS-Tapping.git;
+
+
+        # Build docs
+        - run:
             name: make html
             command: |
               cd doc;
               make html;
-
 
         # Save the outputs
         - store_artifacts:
--- a/.github/workflows/SphinxBuild.yml
+++ b/.github/workflows/SphinxBuild.yml
@@ -15,7 +15,9 @@ jobs:
     - shell: bash -el {0}
       run: |
         cp requirements_doc.txt doc/requirements.txt
-    # Standard drop-in approach that should work for most people.
+    - shell: bash -el {0}
+      run: |
+        git clone --depth 1 https://github.com/rob-luke/BIDS-NIRS-Tapping.git examples/BIDS-NIRS-Tapping
     - uses: ammaraskar/sphinx-action@master
       with:
         docs-folder: "doc/"
--- a/.github/workflows/SphinxBuild_PR.yml
+++ b/.github/workflows/SphinxBuild_PR.yml
@@ -10,6 +10,9 @@ jobs:
     - shell: bash -el {0}
       run: |
         cp requirements_doc.txt doc/requirements.txt
+    - shell: bash -el {0}
+      run: |
+        git clone --depth 1 https://github.com/rob-luke/BIDS-NIRS-Tapping.git examples/BIDS-NIRS-Tapping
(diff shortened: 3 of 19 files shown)


rob-luke:
@larsoner do you have any suggestions how I can get this stats model output to render nicely in this PR example? It looks great in a jupyter notebook, but the text when [here](https://543-254026779-gh.circle-artifacts.com/0/dev/auto_examples/plot_12_group_glm.html#group-results) is all squished. I have also tried rendering as html (which works in jupyter) but that outputs an object rather than the nice table.

Here is a snapshot from the circle ci build...

image

here is what it looks like on my local jupyter notebook...

image

larsoner:
From Chrome inspection:

Screen Shot 2020-09-24 at 7 14 35 AM

So you can style `table.simpletable`, `table.simpletable td` etc. using CSS. If you like the look of [this page]():

Screen Shot 2020-09-24 at 7 16 20 AM

You can look at the CSS used here:

https://github.com/sphinx-gallery/sphinx-gallery/blob/master/sphinx_gallery/_static/gallery-dataframe.css#L4-L36
https://github.com/mne-tools/mne-python/blob/master/doc/_static/style.css#L547-L551

But to get it right I would first inspect the Jupyter output that looks correct, see what style sheet it's coming from, then find which repo has that style sheet. Then maybe open a PR to sphinx-gallery to add their styles as well.

rob-luke:
Thanks @larsoner, I think I can work through that. Appreciate the detailed pointer.

larsoner:
58 commits! Is there a TL;DR for this that would be beneficial (for you) for me to look at? Or some lessons learned? Usually that number of commits means that there was some considerable time spent somewhere...

rob-luke:
Mainly time spent in trying to make an understandable story. I had a few colleagues give me feedback at various points and the feedback was usually the same "make it simpler". For example, I started with a robust linear model but then switched to a mixed effects model as this is more common to people with an R background. I also had in lots of "you could do this, or this or this", but now its a more linear story. Code wise, mostly I learnt about how to slice and edit pandas data frames without throwing warnings using iloc, but that's probably already known to python people. I also spent some time wrangling the different output that stats models produces, they aren't consistent with how the report analysis in tables, so `rlm` returns different columns to `mixedlm`, maybe I'll open a PR there to make consistent one day.

rob-luke:
But its not too late to give feedback on the tutorial, I want to tweak it still and make the API and returned data more consistent. For example there is both `conditions` and `Conditions` used, I want to fix this before tagging a version.

larsoner (commented):
Remarkably short for a group-level analysis!

One high-level idea is that at some point we should see how easily we can integrate with https://github.com/mne-tools/mne-study-template/ .Doing so will ensure BIDS compliance but also that the MNE-study-template is generic enough to support fNIRS

rob-luke:
> Remarkably short for a group-level analysis!

Thanks, its taken a few years to get here! But I am quite happy with the combination of MNE, Nilearn, and StatsModels now.

> see how easily we can integrate with https://github.com/mne-tools/mne-study-template/ .

Yes I like this idea. But I think we hold on a bit before integrating, I have a feeling this approach (and all of MNE-NIRS) will evolve as people find edges cases in it. MNE-NIRS is now perfectly usable, but each study I analyse I find some little quirk. And I guess we want the study template to be semi stable. I expect this repo will be more stable in a few months.

I will create issues from your other points so I don't forget.

PR #172, opened 2020-09-22
WIP: Improve speed of data frame creation

rob-luke:


rob-luke:
This gives me a 8-10x speedup in experiments.

PR #186, opened 2020-10-11
MAINT: Faster build plus memory usage

larsoner:
`mprof run plot_12_group_glm.py` followed by `mprof plot` produces:

Figure_1

Adding `memory_profilier` to your CircleCI build will tell you how much memory examples use. This PR does that, and tweaks the example to use a bit less memory.

Event dropping from 0.6 to 0.5 took it down from ~1.5GB to ~1.2GB, so if that's a better solution feel free to push that change and use it.

### DIFF
--- a/.github/workflows/SphinxBuild.yml
+++ b/.github/workflows/SphinxBuild.yml
@@ -21,7 +21,7 @@ jobs:
     - uses: ammaraskar/sphinx-action@master
       with:
         docs-folder: "doc/"
-        pre-build-command: "apt-get update -y && apt-get install -y python3-psutil"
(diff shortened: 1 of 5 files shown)


rob-luke:
This is fantastic, thanks @larsoner 

Debugging locally is hard, debugging these cloud issues is a real art. Appreciate your help

larsoner:
@rob-luke another magic part:

https://mne.tools/mne-nirs/auto_examples/sg_execution_times.html

The measurements are not always perfect. But you can see that `plot_11_hrf_measured.py` could probably use some trimming.

rob-luke:
That is super cool. Looks like I should downsample that demo too (or get around to investigating the memory sink)

PR #188, opened 2020-10-12
Downsample example to save memory

rob-luke:


larsoner:
Much better, ~1.5GB now

https://646-254026779-gh.circle-artifacts.com/0/dev/auto_examples/sg_execution_times.html

rob-luke:
I think the large memory usage is something I am going to have to address soon. The interface is settling down, so hopefully I can move from making it correct, to making it efficient soon.

Issue #197, opened 2020-11-16
Move unit tests to GitHub actions


rob-luke:
Travis Mac job keeps failing. Move to GitHub actions. See the MNE transition at https://github.com/mne-tools/mne-python/pull/8161

larsoner:
Travis is going to start allowing only 1000 minutes per month for open-source projects. I recommend waiting for https://github.com/mne-tools/mne-python/pull/8514 to land then just move all Travis jobs the same way

rob-luke:
Looks like its landed, ill take a read. Thanks

PR #203, opened 2020-11-23
Use github actions for running tests

rob-luke:
Initial attempt at #197


rob-luke:
Need `lets-plot` to release 3.9 wheel https://pypi.org/project/lets-plot/

Issue #202, opened 2020-11-23
Build docs for multiple versions including dev


rob-luke:
I would like to have the docs available for each released version and for the latest dev branch, the same as MNE.

rob-luke:
Maybe look at https://sphinx-versions.readthedocs.io/en/latest/index.html

rob-luke:
Once this is done then change version to `0.0.2 dev`

rob-luke:
Seems sphinx-versions doesn't work with this theme. Need to find another solution

larsoner:
@rob-luke I would try transitioning completely to pydata-sphinx-theme + sphinx-multiversions See https://github.com/mne-tools/mne-python/issues/8184

rob-luke:
Thanks @larsoner that's exactly what I was after.

rob-luke:
This generally works now, so will close. A list of improvements is listed in #220

PR #207, opened 2020-11-24
Switch to pydata_sphinx_theme

rob-luke:
#202


rob-luke:
This works ok. But sphinx-multiversions didnt work first try, will need to dig further.

PR #211, opened 2020-12-04
Enable topoplotting of subsets of data. Added tests

rob-luke:
Added ability to plot a topo for a single condition and chromophore. And added option to specify vmin, vmax, colorbar. And added tests.


rob-luke:
FYI: @mshader

PR #243, opened 2021-02-12
ENH: Allow OLS in addition to mixedlm

larsoner:
I swapped in `smf.ols` for `smf.mixedlm` and needed to change the `P` key to make it work. Adding a test made me add some other stuff so that `pytest mne_nirs/statistics` ran without emitting any warnings.

### DIFF
--- a/mne_nirs/statistics/tests/test_statsmodels.py
+++ b/mne_nirs/statistics/tests/test_statsmodels.py
@@ -2,6 +2,8 @@
 #
 # License: BSD (3-clause)
 
+import numpy as np
+import pytest
 import pandas as pd
 import statsmodels.formula.api as smf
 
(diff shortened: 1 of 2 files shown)


mne_nirs/statistics/tests/test_statsmodels.py
```diff
@@ -11,7 +13,12 @@
 from ...utils._io import glm_to_tidy
 
 
-def test_statsmodel_to_df():
+@pytest.mark.parametrize('func', ('mixedlm', 'ols'))
```
rob-luke, line 16: Would it be possible to also test rlm here too (I'm away from PC so cant check)?

```suggestion
@pytest.mark.parametrize('func', ('mixedlm', 'ols', 'rlm'))
```

rob-luke:
Awesome, thanks @larsoner

PR #245, opened 2021-02-22
BUG: Fix precision in tables

### LINKED ISSUE #244: BUG: statsmodels_to_results suffers from inaccuracy
Using `.summary()` is nice for printing but bad for numerical computing, so we probably shouldn't use it to get actual values in `statsmodels_to_results`. By way of example:
```
>>> ch_model = smf.mixedlm(
...     "theta ~ -1 + ch_name:Condition",
...     ch_summary, groups=ch_summary["subject"]).fit(method='powell')
>>> ch_model_df = mne_nirs.statistics.statsmodels_to_results(ch_model)
>>> print(np.array(ch_model_df['P>|z|'])[17:33])
[0.    0.135 0.541 0.89  0.146 0.033 0.221 0.19  0.191 0.002 0.112 0.018
 0.129 0.    0.952 0.   ]
>>> print(np.array(ch_model.pvalues)[17:33])
[2.07866441e-07 1.35295118e-01 5.40928010e-01 8.90354390e-01
 1.46297283e-01 3.25360602e-02 2.20535361e-01 1.90387112e-01
 1.91051924e-01 2.39444468e-03 1.11936842e-01 1.77914955e-02
 1.29375584e-01 1.17267783e-04 9.52492330e-01 5.13254033e-05]
```
I think in the current code, any p-value less than `0.0005` significance will becomes `0.0` due to imprecision.

larsoner:
Some of the quantities are derived so I had to adapt code from statsmodels (yuck). But I think it works at least.

Closes #244

### DIFF
--- a/mne_nirs/statistics/_statsmodels.py
+++ b/mne_nirs/statistics/_statsmodels.py
@@ -3,10 +3,15 @@
 # License: BSD (3-clause)
 
 import pandas as pd
+import numpy as np
 
 
 def summary_to_dataframe(summary):
-    '''Convert statsmodels summary to pandas dataframe'''
+    '''Convert statsmodels summary to pandas dataframe.
+
+    .. warning:: The summary has precision issues, use the numerical values
+                 from it with caution.
+    '''
     results = summary.tables[1]
     if type(results) is not pd.core.frame.DataFrame:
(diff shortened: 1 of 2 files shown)


rob-luke:
Brilliant, thanks @larsoner

PR #246, opened 2021-02-25
Add ignore_words to spell check

rob-luke:


rob-luke:
Addressing broken circle ci in #248

PR #249, opened 2021-02-25
General improvements to tests

rob-luke:


mne_nirs/channels/tests/test_roi.py
```diff
@@ -32,9 +32,10 @@ def test_roi_picks():
     with pytest.raises(ValueError, match='No matching'):
         picks_pair_to_idx(raw, [[1, 1], [1, 2], [15, 13], [8, 16]])
 
-    picks = picks_pair_to_idx(raw, [[1, 1], [1, 2], [15, 13], [8, 16]],
-                              on_missing='warning')
-    assert len(picks) == 6  # Missing should be ignored
+    with pytest.warns(RuntimeWarning, match='No matching channels'):
+        picks = picks_pair_to_idx(raw, [[1, 1], [1, 2], [15, 13], [8, 16]],
+                                on_missing='warning')
+        assert len(picks) == 6  # Missing should be ignored
```
larsoner, line 38: Nitpicks -- missed indent, and best only to have the thing that emits the warning in the context manager (more target)
```suggestion
                                  on_missing='warning')
    assert len(picks) == 6  # Missing should be ignored
```

rob-luke:
That makes sense @larsoner.

And I don't think of it as nitpicking, but as personalised programming coaching.

PR #280, opened 2021-06-06
WIP: Add peak spectral power metric

### LINKED ISSUE #278: ENH: Add automated bad-channel marking
I recently migrated a pipeline from HOMER to MNE-Python. The one function I was missing was one that marked a channel in a recording if the optical density was over 3 standard deviations from the group mean (I think this was the criteria anyway, and it seems to match what they suggest offhand in [their publication](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2761652/#S11title)). Is it worth implementing something like this here? If so I have a bunch of recordings I can validate against. If not, I'm curious to know what you use instead @rob-luke !

It's not 100% clear to me, but I could see this being implemented as either 1) 3 std devs away from the mean *across all channels*, or 2) operating based on the std/mean of *each channel* independently -- probably the latter makes more sense. Then we'd want to use channel-specific annotations from MNE to mark segments as bad for segments of time during which it exceeded the threshold. Then users can choose to mark the entire channel bad for a given file/run (which the previous analysis I have access to did) or to have those segments marked bad in time. However, this latter method will be a bit tricky because in principle I think we'd want to exclude that interval from the GLM somehow -- which seems hard to do for the entire recording, let alone just for a single channel.

Mostly thinking out loud here @rob-luke, perhaps it makes sense to have a quick chat about this at some point!

rob-luke:
Adds the peak spectral power metric [1, 2]. Closes #278 

[1] Pollonini L et al., “PHOEBE: a method for real time mapping of optodes-scalp coupling in functional near-infrared spectroscopy” in Biomed. Opt. Express 7, 5104-5119 (2016). 
[2] Hernandez, Samuel Montero, and Luca Pollonini. "NIRSplot: a tool for quality assessment of fNIRS scans." Optics and the Brain. Optical Society of America, 2020.


rob-luke:
Here are some figures from real measurements. The left plot of the raw values is informative. The right figure shows the selection of bad channels/times by thresholding the scores on the left. It seems like the threshold isn't catching all the sections I would choose. I will dig in to the code to make sure I dont have a scaling issue somewhere, and read papers to find out if the threshold is always the same, or seems dependent on the study.

image

image

image

rob-luke:
@larsoner could you please review. Does the combination of SCI and PP meet your requirements?I tried to keep the plotting function generic so we could also implement the std approach you suggested.

rob-luke:
@mshader fyi and review too if you want.

mne_nirs/preprocessing/_peak_power.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    Compute peak spectral power metric from [1, 2].
+
+    [1] Pollonini L et al., “PHOEBE: a method for real time mapping of
+    optodes-scalp coupling in functional near-infrared spectroscopy” in
+    Biomed. Opt. Express 7, 5104-5119 (2016).
+
+    [2] Hernandez, Samuel Montero, and Luca Pollonini. "NIRSplot: a tool for
+    quality assessment of fNIRS scans." Optics and the Brain.
+    Optical Society of America, 2020.
```
larsoner, line 30: Typically these go in a `References` section at the end of the docs, and are referred to inline via `... from [1]_, [2]_.` or so

mne_nirs/preprocessing/_peak_power.py
```diff
@@ -0,0 +1,104 @@
...
+        The duration of the window over which to calculate the metric.
+        Default is 10 seconds as in PHOEBE paper.
+    threshold : number
+        Values below this are marked as bad and annotated in the raw file.
+    %(l_freq)s
+    %(h_freq)s
+    %(l_trans_bandwidth)s
+    %(h_trans_bandwidth)s
+    %(verbose)s
+    Returns
```
larsoner, line 46: You need a newline before this Returns or it won't render properly:

doc/api.rst
```diff
@@ -24,6 +24,21 @@ Experimental Design and Analysis
...
+
+.. automodule:: mne_nirs.preprocessing
+   :no-members:
+   :no-inherited-members:
+
+.. autosummary::
+   :toctree: generated/
+
+   peak_power
+
```
larsoner, line 40: I am confused how/why this does not show up in the docs rendered by CircleCI:

https://1130-254026779-gh.circle-artifacts.com/0/dev/master/api.html

Looking at the CircleCI logs I don't see the example you added being run, either (and it's not in the artifacts):

https://app.circleci.com/pipelines/github/mne-tools/mne-nirs/1122/workflows/6ca1b2f3-0720-4482-b8db-21da827113e9/jobs/1130

larsoner (commented):
> I find the threshold could be set a big higher at 0.3-0.5. But this may be due to the fact that I spend LOTS of time to ensure my data is clear during collection. So I leave the threshold as the default of 0.1 as suggested by the method author in original paper.

Some variant of this in Notes would be helpful for people

Once the CircleCI rendering is fixed I'll take a look at the example (it's much easier/nicer than looking at the RST directly)

rob-luke:
@larsoner the rendering is being run correctly. See the links in the PR description above

doc/api.rst
rob-luke: I'm guessing you were looking at master rather than the pull branch. You need to select what you want to view in the top right drop down.

larsoner: Ahh yes I didn't even notice that it built multiple versions of the docs!

mne_nirs/preprocessing/_peak_power.py
larsoner: Indeed it does not currently render properly:

https://1130-254026779-gh.circle-artifacts.com/0/dev/pull/280/generated/mne_nirs.preprocessing.peak_power.html#mne-nirs-preprocessing-peak-power

doc/api.rst
rob-luke: I'd like to make the pull version the default when clicking through the CI. But haven't got around to it.

larsoner: I would add a conditional when building on CircleCI that if it's a PR, you add a tiny redirect HTML that points to the correct URL (since it's PR-dependent, there are CircleCI vars you can use for this), then tell the artifacts-redirector-action to default to this redirect HTML file's URL

larsoner:
I'll take a look at the example tomorrow morning (~10 h from now)!

rob-luke:
No rush. I also invited a review via email. So might not merge till that comes back.

larsoner:
Example looks useful to me, I'll give this in a try in the next week or two, but don't wait for that to merge. In the meantime there are some suggestions above about formatting refs and sections, etc. I'd also recommend adding this to `conf.py`, it would have caught the `Returns` section problem:

https://github.com/mne-tools/mne-python/blob/0d5068dbbce76598c9c6433e1ace1265e1955564/doc/conf.py#L255-L269

PR #287, opened 2021-06-09
WIP: Plot GLM results on cortical surface

### LINKED ISSUE #285: Add ability to interface stc_to_sensor with GLM results

rob-luke:
closes #285 
Just a proof of concept now. Need to figure out the API etc

Initial render looks good https://1175-254026779-gh.circle-artifacts.com/0/dev/pull/287/auto_examples/general/plot_12_group_glm.html#cortical-surface-projections

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -92,7 +92,6 @@ Visualisation
    plot_glm_group_topo
    plot_timechannel_quality_metric
    plot_glm_surface_projection
-   brain_to_image
 
 
 Simulation
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -91,6 +91,8 @@ Visualisation
    plot_glm_contrast_topo
    plot_glm_group_topo
    plot_timechannel_quality_metric
+   plot_glm_surface_projection
+   brain_to_image
 
 
 Simulation
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -223,6 +223,7 @@ intersphinx_mapping = {
 sphinx_gallery_conf = {
     'doc_module': 'mne_nirs',
     'backreferences_dir': os.path.join('generated'),
+    'image_scrapers': ('pyvista', 'matplotlib'),
     'reference_url': {
         'mne_nirs': None},
     'download_all_examples': False,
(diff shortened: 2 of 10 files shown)


rob-luke:
Hi @larsoner would you be able to check if my understanding of the difference between sum and weighted is described here correctly? I think I will only show one of these modes in the final version of the tutorial, but I want to make sure I understand things correctly before choosing a default. Thanks

https://1206-254026779-gh.circle-artifacts.com/0/dev/pull/287/auto_examples/general/plot_12_group_glm.html#id9

larsoner:
Yes that sounds right. I actually regret adding "sum" mode, I don't think it's really all that useful -- "weighted" seems like the sensible choice.

> I should read some fMRI literature and see which figure best corresponds to what would be expected.

I don't really think they deal with this problem of projecting sensors. But at least in interpolation schemes I think typically the weighting is kept normalized like "weighted".

rob-luke:
Awesome I have set the default to weighted then.

@larsoner I have another problem I cant seem to solve (you can probably tell from the commit list above ☹️ ). The new plots render nicely on CircleCI (see https://1246-254026779-gh.circle-artifacts.com/0/dev/pull/287/auto_examples/general/plot_12_group_glm.html#cortical-surface-projections). But the github actions generated docs just shows a black box with a Colorbar, this is what it looks like...

sphx_glr_plot_12_group_glm_004

Do you have any suggestions as to what may be causing this? I tried using different 3D backend (mayavi). Are there any other obvious candidates for the error?

I know that MNE-Python doesnt use github actions for docs, so I am not asking you to solve the issue, but I hoped this might be a common issue that you’ve seen before with known solution 😉 .

larsoner:
My first thought was that it could be a Mesa GL version issue, but the info shows `Mesa 20.2.6`, which is plenty new enough. CircleCI appears to use 18.3.6, which is older (from `buster` debain) -- maybe for some reason it actually works better?

I would stick with the PyVista backend since it's the most well tested. Other than the above Mesa version issue somehow being a problem, I'm not sure why it wouldn't render properly :(

rob-luke:
I have been unable to figure out how to generate 3D brain figures using GitHub Actions (see #297 ). As such, I will demonstrate this functionality but make a note in the docs that it will only work locally.

PR #303, opened 2021-07-02
Create a class for storing GLM results

### LINKED ISSUE #304: Change function run_GLM to run_glm
According to https://www.python.org/dev/peps/pep-0008/#function-and-variable-names 

"Function names should be lowercase, with words separated by underscores as necessary to improve readability."

rob-luke:
This PR aims to simplify the user experience by adding types for storing the GLM results.

Currently the user needs to pass around the Info, GLM results, and design matrix. This PR creates a class to handle this. It also provides useful functions.

This PR adds hooks to existing functionality. In follow up PRs I will add more class functions such as plotting, saving and loading data, etc.

This PR:
* Adds a class for glm regression results. Its just a wrapper containing the nilearn regression results and the mne info structure.
* Adds a class for glm contrast results. Similar to above its a wrapper.
* Each class has several methods such as plot_topo, to_dataframe etc. That are just new interface to the existing code
* Depreciates several of the existing functions in favour of the classes
* Demonstrates the new classes in the examples

All previous functions still exist, but will be removed in v1.0.0 which would be an API breaking version.

Closes #304


rob-luke:
@larsoner this is my first time writing a class in python. Could you please provide an early review to let me know if I am on the right path or not? Have I made any rookie mistakes :) The class is defined in `mne_nirs/statistics/_glm_type.py`.

I need to add more tests and update the tutorial wording etc, but I think the tutorial code is already cleaner. Also please be critical of function names, I'd like to stabilise the API for a while after this.

doc/changelog.rst
```diff
@@ -12,6 +12,15 @@ Major version changes indicate incompatible API changes.
 Minor version changes indicate new functionality was added in a backwards compatible manner.
 Patch version changes indicate backward compatible bug fixes.
 
+v0.1.0
+------
+
+Enhancements
+
+* Add RegressionResults and ContrastResults classes to store GLM results. These classes have methods that replace the functions compute_contrast and glm_region_of_interest. By `Robert Luke`_.
```
larsoner, line 20:
```suggestion
* Add :class:`~mne_nirs.statistics.RegressionResults` and :class:`~mne_nirs.statistics.ContrastResults` classes to store GLM results. These classes have methods that replace the functions compute_contrast and glm_region_of_interest. By `Robert Luke`_.
```

doc/conf.py
```diff
@@ -25,11 +25,14 @@
...
 
 
+# -- Path setup --------------------------------------------------------------
 
 # If extensions (or modules to document with autodoc) are in another directory,
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
-# sys.path.insert(0, os.path.abspath('.'))
+curdir = os.path.dirname(__file__)
+sys.path.append(os.path.abspath(os.path.join(curdir, '..', 'mne_nirs')))
```
larsoner, line 34: this shouldn't really be necessary for any dev with a sane doc build setup (e.g., installed with `python setup.py develop` or `pip install -ve.`

doc/changelog.rst
larsoner: ... but I think it's important to tell people that an important side effect of this change is that a return type has changed from dataframe to these classes, and that if they want they can do `.to_dataframe()` or whatever to get what they had in <= v0.0.6 back. (This assumes that it does indeed give back an otherwise identical dataframe.) This will save folks like me who have existing code that operates directly on the dataframe some pain in transition.

mne_nirs/statistics/_glm_level_first.py
```diff
@@ -0,0 +1,580 @@
...
+        'all CPUs'.
+    verbose : int, optional
+        The verbosity level. Default is 0.
+
+    Returns
+    -------
+    glm_estimates : RegressionResults
+        RegressionResults class which stores the GLM results.
+    """
+    from nilearn.glm.first_level import run_glm
```
larsoner, line 527: You do unnested nilearn imports at the top, so no reason to nest this I think?

larsoner (commented):
I didn't look too deeply at the implementations but generally at the class structure. It all looks reasonable to me as ways of simplifying examples and common use cases.

The one thing that I expected to see but did not is some form of object persistence like `obj.save(...)` and `mne_nirs.statistics.read_glm(...)`. If you look in MNE we do it very easily via `mne.externals.h5io.write_hdf5` and `mne.externals.h5io.read_hdf5` which already support dataframes and info objects etc., so I think it should be pretty easy to add it if you want to here or in another PR.

rob-luke:
Huge thanks @larsoner. That high level check gives me exactly the confidence I need to push on and add new features like save and load. Cheers!

doc/conf.py
rob-luke: Thanks, I will remove. This was a remnant from me trying o fix the numpydoc error I am getting, I've tried like 50 different things. Grrr

mne_nirs/statistics/_glm_level_first.py
rob-luke: Yes, thanks for the keen eyes!

doc/changelog.rst
rob-luke: I agree with you. I think I will write an extended changelog for this version based on your suggestion.

PR #305, opened 2021-07-02
MAINT: Update for Hitachi

larsoner:
Use newer method for verifying freqs to avoid:
```
  File "<decorator-gen-528>", line 22, in peak_power
  File "/home/larsoner/python/mne-nirs/mne_nirs/preprocessing/_peak_power.py", line 65, in peak_power
    picks = _check_channels_ordered(raw.info, freqs)
  File "/home/larsoner/python/mne-python/mne/preprocessing/nirs/nirs.py", line 159, in _check_channels_ordered
    raise ValueError(
ValueError: NIRS channels not ordered correctly. Channels must be ordered as source detector pairs with alternating frequencies: 698 & 699
```

PR #306, opened 2021-07-02
Update CircleCI apt

rob-luke:
From https://github.com/mne-tools/mne-nirs/pull/305/files


rob-luke:
Ops I got mixed up between repos

PR #313, opened 2021-07-05
Update changelog.rst

rob-luke:
Token change to try and debug why the circle ci link fails

Note: in an attempt to debug this issue I have opened it from the organisation repo rather than my own fork.

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -31,7 +31,7 @@ Enhancements
 
 * Added windowed signal quality metrics scalp coupling index and peak power. By `Robert Luke`_.
 
-* Added FIR GLM example. By `Robert Luke`_.
+* Added a finite impulse response (FIR) GLM example. By `Robert Luke`_.
 
 * Added group level waveform example. By `Robert Luke`_.
 
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -27,14 +27,13 @@ Enhancements
 v0.0.5
 ------
 
-* Added windowed signal quality metrics scalp coupling index and peak power.
+* Added windowed signal quality metrics scalp coupling index and peak power. By `Robert Luke`_.
 
-* Added FIR GLM example.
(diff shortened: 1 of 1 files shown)


rob-luke:
@larsoner could you provide any suggestion as to why my redirect to CircleCI doesn't work?

The GitHub action is located at https://github.com/mne-tools/mne-nirs/blob/master/.github/workflows/CircleCI_PR.yml

As your page for this action says, its a bit hard to debug, and because im not admin I cant set `ACTIONS_STEP_DEBUG=true` as you suggest.

The link always points to `0/dev/pull//index.html` but I would like it to point to `0/dev/pull/${{ env.PR_NUMBER }}/index.html`

Thanks

larsoner:
Looking at this:

https://github.com/mne-tools/mne-nirs/runs/2988196521?check_suite_focus=true

You can see that the env var is incorrect:
```
    artifact-path: 0/dev/pull//index.html
```
And debugging you see that it's not correct even before this:
```
Run echo "PR number = "
  echo "PR number = "
  echo "PR number = "
  shell: /usr/bin/bash -e {0}
  env:
    PR_NUMBER: 
```
This is because the `event` that is triggered here is [status](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#status) not `pull_request`. So somehow you have to get it from somewhere else within the [status payload](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#status) or use something from there (like the SHA) to figure out what PR it's a part of (if anything). For example you could look at:

https://developer.github.com/v3/repos/commits/#list-pull-requests-associated-with-commit

Or:

https://stackoverflow.com/questions/59077079/how-to-get-pull-request-number-within-github-actions-workflow#answer-60723400

rob-luke:
Thanks @larsoner for the detailed explanation. That makes sense to me, I hadn't realised about the status workflow. Your direction should help get me on the right path when I have a moment.

PR #326, opened 2021-07-15
Fix snirf file inconsistencies with the spec

### LINKED ISSUE #317: Fix inconsistencies between written SNIRF files and the SNIRF spec
#### Describe the bug
hdf5 files written via `mne_nirs.io.snirf.write_raw_snirf` have a few inconsistencies with the (SNIRF File Format spec)[https://github.com/fNIRS/snirf/blob/master/snirf_specification.md].

In particular, the ones I've noticed so far are:
* required `/formatVersion` is missing
* required `/nirs/metaDataTags/TimeUnit` is missing
* required `/nirs/metadataTags/FrequencyUnit` is missing
* required `/nirs/data(i)/measurementList(i)/dataType` is only present for `measurementList1`
* required `/nirs/data(i)/measurementList(i)/dataTypeIndex` is missing
* `/nirs/stim(i)` should be using integer i values rather than floats (eg. the current files have `stim1.0` instead of `stim1`)
* required `/nirs/stim(i)/name` is missing

#### Steps to reproduce
Explore the snirf file using h5py or an HDF explorer like HDFView and compare against the (SNIRF File Format spec)[https://github.com/fNIRS/snirf/blob/master/snirf_specification.md].

#### Expected results
That the output snirf files are consistent with the SNIRF format spec.

#### Actual results
The inconsistencies listed above are present.

#### Additional information
Are any of these inconsistencies intentional?

dsleiter:
Fix #317

This includes a fix for the inconsistencies between the snirf writer and the [snirf spec (Version 1.0 - Draft 3)](https://github.com/fNIRS/snirf/blob/master/snirf_specification.md#nirsidatajmeasurementlistk).

@rob-luke I will mark this as a draft PR because I have a few questions:

1. For the formatVersion it wasn't clear to me if there was a decision on version numbering prior to draft 4 when implemented would become v 1.0. So for now I've labeled it "1.0 - Draft 3". While that doesn't fit a clear convention, at least it should uniquely identify the spec, and can be updated to 1.0 once draft 4 is out. Is that ok for now, or would you prefer a different format?
1. I wasn't able to find an explicit time or frequency unit in the Raw object. Is it there and I missed it? Is there any specification on what the time or frequency units should be? For now I used 's' and 'Hz'.
1. the significance of dataTypeIndex isn't clear when using CW Amplitude. However, the spec gave an example where it set dataType=1 and dataTypeIndex=1 for CW Amplitude, so I copied that here.
https://github.com/fNIRS/snirf/blob/master/snirf_specification.md#nirsidatajmeasurementlistk
(after the moduleIndex sub-section)
1. SNIRF stim groups are now indexed sequentially from 1 to N (instead of using the description). The stim groups have a required 'name' parameter, while the only text component of the Annotations is 'description'. While it isn't an exact conceptual match, the description for the SNIRF 'name' field says: "This is a string describing the jth stimulus condition." Therefore I went ahead and mapped the Annotation description to the SNIRF stim name. Does that seem reasonable enough to you?

Finally, this includes a breaking change in the name of the stim groups: previously stim groups were named "stim1.0", "stim2.0", stim15.0", etc (not necessarily sequential), they are now sequentially named "stim1", "stim2", etc. I believe that the previous naming was incompatible with the spec so this is a bug fix, but this will be a breaking change for anyone relying on the previous naming convention.

Since the tests for the snirf writer rely on the MNE snirf reader which uses the existing stim group naming, the existing test breaks. I will submit a PR to MNE-Python as well (I already have a minimal fix locally), but we still need tests to pass here before we can merge this PR. Would you like to wait until the MNE PR is accepted and a new release is made? Or can we comment out the [annotations description check in `test_snirf_write`](https://github.com/mne-tools/mne-nirs/blob/832318363fb4e8a2002ae2cb82869d4522d41b08/mne_nirs/io/snirf/tests/test_snirf.py#L40-L41) for now (and create an issue to uncomment the annotations check once the next MNE version is released with the fix)?

### DIFF
--- a/mne_nirs/io/snirf/_snirf.py
+++ b/mne_nirs/io/snirf/_snirf.py
@@ -21,17 +21,17 @@ def write_raw_snirf(raw, fname):
         Path to the SNIRF data file.
     """
 
-    picks = _picks_to_idx(raw.info, "fnirs_cw_amplitude", exclude=[])
-    assert len(picks) == len(raw.ch_names), "Data must be fnirs_cw_amplitude"
+    picks = _picks_to_idx(raw.info, 'fnirs_cw_amplitude', exclude=[])
+    assert len(picks) == len(raw.ch_names), 'Data must be fnirs_cw_amplitude'
 
     # Reordering channels
     num_chans = len(raw.ch_names)
     raw = raw.copy()
     raw.pick(picks=list(range(num_chans)[0::2]) + list(range(num_chans)[1::2]))
 
-    with h5py.File(fname, "w") as f:
-        nirs = f.create_group("/nirs")
-        f.create_dataset("formatVersion", data=[_str_encode("1.0")])
+    with h5py.File(fname, 'w') as f:
+        nirs = f.create_group('/nirs')
+        f.create_dataset('formatVersion', data=[_str_encode('1.0')])
 
         _add_metadata_tags(raw, nirs)
         _add_single_data_block(raw, nirs)
@@ -47,7 +47,7 @@ def _str_encode(str_val):
     str_val : str
         The string to encode.
     """
-    return str_val.encode("UTF-8")
+    return str_val.encode('UTF-8')
 
 
 def _add_metadata_tags(raw, nirs):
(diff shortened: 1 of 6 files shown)


dsleiter:
Here is the related MNE-Python issue: https://github.com/mne-tools/mne-python/issues/9571

dsleiter:
@rob-luke the `test_snirf_write` test is currently failing, but this is expected as it uses the MNE `read_raw_snirf` function, which won't hasn't been updated to match the change to the Annotation description.

It isn't clear to me why the coverage drops so significantly or why the numpydoc CI fails. Do you have any input on those?

rob-luke:
Thanks @dsleiter. 

> why the numpydoc CI fails

This isn't caused by your PR. This was happening before, I just haven't been able to figure out how to fix it. Please ignore this failure.

> why the coverage drops so significantly

Agreed, this is strange. Maybe its a bug with the CI. Once the MNE-Python PR is merged we will run all the CIs here again and hope it magically fixes itself.

> this is expected as it uses the MNE read_raw_snirf function, which won't hasn't been updated to match the change to the Annotation description.

That makes sense. I will pull both branches and test locally, then report back.

rob-luke:
>  I've labeled it "1.0 - Draft 3".

Yeah I agree its odd, but I think my attempts to convince them otherwise have failed. Lets use this.

> I wasn't able to find an explicit time or frequency unit in the Raw object. Is it there and I missed it? Is there any specification on what the time or frequency units should be? For now I used 's' and 'Hz'.

I think everything should be SI units. So that's right.

> I went ahead and mapped the Annotation description to the SNIRF stim name

Sounds great.

-------

Ok this passed locally for me when using the appropriate MNE-Python pr branch. I also included a real measurement in the tests which ensures the data is retained when converting from NIRx vendor format to snirf.

I'm happy with this. Lets see if we can get a review from @larsoner or @agramfort too.

rob-luke:
> Would you like to wait until the MNE PR is accepted and a new release is made? Or can we comment out the annotations description check in test_snirf_write for now 

Yes lets wait for the MNE-Python PR to be merged, but not wait for a new release (could be months, I dont know the schedule). We will use MNE-Python main for testing the MNE-NIRS code, so shouldn't be a problem.

larsoner:
> Yes lets wait for the MNE-Python PR to be merged,

Done!

rob-luke:
@dsleiter I just made a trivial change to the README to make the CIs kick off again. Hopefully everything will come back green if its using the new merged MNE-Python code you submitted.

rob-luke:
I am going to need to remake my example demo set with your new improved SNIRF. Ill do that, upload it to GitHub, then kick these CIs off again. Cant do it now, give me a day 😉

mne_nirs/io/snirf/_snirf.py
```diff
@@ -20,138 +22,311 @@ def write_raw_snirf(raw, fname):
...
-            f.create_dataset("nirs/metaDataTags/middleName", data=mname)
-        if 'last_name' in raw.info["subject_info"]:
-            lname = [raw.info["subject_info"]['last_name'].encode('UTF-8')]
-            f.create_dataset("nirs/metaDataTags/lastName", data=lname)
-        if 'sex' in raw.info["subject_info"]:
-            sex = str(int(raw.info["subject_info"]['sex'])).encode('UTF-8')
-            f.create_dataset("nirs/metaDataTags/sex", data=[sex])
+    with h5py.File(fname, 'w') as f:
+        nirs = f.create_group('/nirs')
+        f.create_dataset('formatVersion', data=[_str_encode('1.0')])
```
rob-luke, line 34: > So for now I've labeled it "1.0 - Draft 3". 

I think it's currently just "1.0". Im ok with either option, your call.

mne_nirs/io/snirf/tests/test_snirf.py
```diff
@@ -56,3 +58,58 @@ def test_snirf_write(fname, tmpdir):
             # logno and scanno are not used in processing
             diffs += f'\n{line}'
     assert diffs == ''
+
+    _verify_snirf_required_fields(test_file)
+
+
+def _verify_snirf_required_fields(test_file):
```
rob-luke, line 65: This is really nice. In the future we could make it a public function as a SNIRF validator.

rob-luke (approved):
This looks great @dsleiter. I updated the test dataset and now the CIs are green. Could you please add an entry to `doc/changelog.rst`, then I will hit merge if you are happy.

(sorry, I know the PR process wasn't documented and nowhere tells you to update the changelog).

mne_nirs/io/snirf/_snirf.py
dsleiter: Good catch! I initially put 1.0, and then thought including the draft number would be better because it is explicit, but forgot to change the code.

I still think "1.0 - Draft 3" would be better, so I'll make an update. Will you need to make a new test dataset and update the hash?

rob-luke: I dont think that will break the reader, so shouldn't require a new dataset. But if I am wrong its easy for me to do. Push away and ping me when you want me to hit merge.

mne_nirs/io/snirf/tests/test_snirf.py
dsleiter: Sounds good! I think that could be beneficial. I tried out the [SNIRF validator from the SNIRF github](https://github.com/BUNPC/python_snirf_validator/tree/66a6e65aad66108812354825a4835280110ca565) without success, but looks like a lot of the work has already been done there. Maybe we can combine the two.

rob-luke: Yes I have had no luck with it either. And agree, best to combine efforts.

dsleiter:
@rob-luke ok, version string and changelog updated. I changed it from draft to ready to merge, so assuming CI goes well I think it should be ready to go! Thanks for the review!

rob-luke:
Thanks @dsleiter

PR #327, opened 2021-07-18
Add example of how to include second level covariates in group level GLM tutorial

### LINKED ISSUE #301: Add example of how to include covariates in second level statistics
For example age as extracted from `raw.info["subject_info"]`

rob-luke:
Closes #301


larsoner:
Always a sign of a nicely extendable API where something like this requires more explanatory text than lines of new code, and no new code in the module itself. Thanks statsmodels!

rob-luke:
Yes the Rogers Wilkinson notations is very powerful. And the great API from stats models is why I have been hesitant to create my own group level class, I don't want to remove any of the flexibility that stats models provides. But I think I will eventually make a group level GLM Results class to simplify plotting etc.

PR #336, opened 2021-08-02
Test multiple pip versions

rob-luke:


rob-luke:
Fails pre 3.9 due to 

https://github.com/mne-tools/mne-python/blob/8d8609ade86b6daca5173faa7414885bde04df29/tools/github_actions_dependencies.sh#L24-L26

So I am not going to test earlier either

PR #337, opened 2021-08-02
Add glm.save and read_glm

### LINKED ISSUE #116: Demonstrate how to save analysis and resume
> Do you want to update a MNE-nirs example to show how you envision users using this API / saving to get back to an analysis / coefficients that they previously computed? https://github.com/mne-tools/mne-nirs/pull/111#issuecomment-643837608

with some greater pointers from larsoner here  https://github.com/mne-tools/mne-nirs/pull/111#issuecomment-643613639

rob-luke:
Closes #116

Implement a method to write GLM results to disk and read them back.
Handles the wrangling of nested types.

- [x] Confirm the approach taken is reasonable
- [x] ~~Is there a way to test if two objects are the same? Currently I test if the resulting dataframe is identical.~~ I added an `__eq__` method.
- [x] Also implement for ContrastResults
- [x] Add more tests specifically checking types:
  - [x] ContrastResults, OLSModel
- [x] Add tutorial

### DIFF
--- a/doc/api.rst
+++ b/doc/api.rst
@@ -168,6 +168,7 @@ Individual (first) level result classes.
 
    RegressionResults
    ContrastResults
+   read_glm
    
 Individual (first) level result class methods.
 View the class documentation above for a detailed list of methods.
@@ -189,10 +190,12 @@ View the class documentation above for a detailed list of methods.
    RegressionResults.plot_topo
    RegressionResults.surface_projection
    RegressionResults.scatter
+   RegressionResults.save
    
    ContrastResults.to_dataframe
    ContrastResults.plot_topo
    ContrastResults.scatter
+   ContrastResults.save
 
 
 Second level analysis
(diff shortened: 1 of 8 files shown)


rob-luke:
@larsoner could you please give this a review? (The numpydoc failure is existing issue). There is quite some data wrangling, so maybe an easier solution exists.

rob-luke:
Here is the tutorial render as i never solved the linking issue https://1580-254026779-gh.circle-artifacts.com/0/dev/pull/337/auto_examples/general/plot_80_save_read_glm.html#sphx-glr-auto-examples-general-plot-80-save-read-glm-py

larsoner:
The approach seems reasonable to me, we do something similar for saving and loading dipy classes in MNE-Python

rob-luke:
Appreciate you looking @larsoner. I'll check out the dipy code you mention. And then push on with this.

larsoner:
That code lives in mne/morph.py. we also use a state-based approach in csd.py I think

Issue #348, opened 2021-08-11
SNIRF writer should not require birthdate


rob-luke:
#### Describe the bug
The SNIRF writer is requiring a birthdate. But this is only a recommended field for SNIRF.

#### Steps to reproduce
Load a file without birthdate, then try to write it back to disk.

#### Expected results
It writes to disk

#### Actual results
It crashes and points to lines https://github.com/mne-tools/mne-nirs/blob/341a2f38a7bef1899f473db52eda4197627c9c81/mne_nirs/io/snirf/_snirf.py#L74-L80

rob-luke:
Closed in #349

PR #349, opened 2021-08-13
Birth Date Fix for SNIRF Files

chrsthur:
- Fixed metadata parameter requirements such that birth date is now optional

Linked Issue: SNIRF writer should not require birthdate #348

### DIFF
--- a/mne_nirs/io/snirf/_snirf.py
+++ b/mne_nirs/io/snirf/_snirf.py
@@ -89,7 +89,8 @@ def _add_metadata_tags(raw, nirs):
     if 'birthday' in raw.info['subject_info']:
         birthday = datetime.date(*raw.info['subject_info']['birthday'])
         birthstr = birthday.strftime('%Y-%m-%d')
-        metadata_tags.create_dataset('DateOfBirth', data=[_str_encode(birthstr)])
+        metadata_tags.create_dataset('DateOfBirth',
+                                     data=[_str_encode(birthstr)])
     if 'middle_name' in raw.info['subject_info']:
         middle_name = raw.info['subject_info']['middle_name']
         metadata_tags.create_dataset('middleName',
--- a/mne_nirs/io/snirf/_snirf.py
+++ b/mne_nirs/io/snirf/_snirf.py
@@ -68,16 +68,13 @@ def _add_metadata_tags(raw, nirs):
     """
     metadata_tags = nirs.create_group('metaDataTags')
 
-    # Store measurement and birth date
+    # Store measurement
     datestr = raw.info['meas_date'].strftime('%Y-%m-%d')
     timestr = raw.info['meas_date'].strftime('%H:%M:%SZ')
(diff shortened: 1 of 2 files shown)


rob-luke:
Hi @chrsthur, thanks for contributing this fix! Its greatly appreciated.

As a first time contributor GitHub will require me to authorise the CI runs. So feel free to @ my name if I need to click ok again.

Initial check of the code looks good. Lets see what the CIs say. Please ignore the numpydoc check, that's broken. The rest should come back with green ticks. If they dont, you can click on the details button and it will tell you what the error was.

Thanks again. Also FYI @dsleiter

rob-luke:
@chrsthur I pushed a commit to your branch that added a test. This will ensure that we don't accidentally revert your change in the future. The only other thing was the lint check is failing. Otherwise looks good. Ping me when the style checks and test are green. Thanks

rob-luke:
All green, thanks @chrsthur

rob-luke:
And I have cut a new release of https://github.com/rob-luke/fnirs-apps-sourcedata2bids/releases/tag/v0.4.1

dsleiter:
Looks great! Thanks, @chrsthur!

@rob-luke I did a quick review to see if there were any other optional SNIRF fields that the writer requires, and found two other potential cases, both for the probe.

1. `/nirs{i}/probe/sourceLabels` and `/nirs{i}/probe/detectorLabels` are optional. Currently those labels are parsed from `mne.io.Raw.ch_names`. Is there a possibility of channel names not existing in a `Raw` object containing fNIRS data?
2. the writer currently writes `/nirs{i}/probe/sourcePos3D` and `/nirs{i}/probe/detectorPos3D` and doesn't handle 2D positions. Is there ever a case where an `mne.io.Raw` fNIRS object would have 2D rather than 3D positions?

On a separate note: the SNIRF format says that the 2D positions are required while the 3D positions are optional. I had assumed that would be either/or situation, but perhaps it should be writing out 2D positions always (alongside the 3D positions if they exist). Do you think it is worth adding the 2D positions?

rob-luke:
Thanks for checking @dsleiter 

> . Is there a possibility of channel names not existing in a Raw object containing fNIRS data

Not possible, so we are safe here

> Is there ever a case where an mne.io.Raw fNIRS object would have 2D rather than 3D positions?

Yeah the 2D requirement is because all the established labs used 2D positions. Currently if only 2d positions are available in the raw data, MNE will still store it as 3D internally and set z=0. So we could check if all(z==0) and then write to the 2D field in SNIRF. But I think we can just tackle that hurdle when we need to. If someone is only using 2D positions then they probably arent that serious about locations anyway.

I will raise this 2d requirment in the SNIRF repo, as i think its a bit odd

dsleiter:
Ok, great. SNIRF might have a legacy reason for keeping the 2D positions around, but it would certainly be helpful if it was either/or.

PR #355, opened 2021-08-16
MAINT: Restore numpydoc

larsoner:
Follow-up to #351


.github/workflows/CircleCI_PR.yml
```diff
@@ -1,4 +1,7 @@
 on: [status]
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.number }}
+  cancel-in-progress: true
```
larsoner, line 4: I also added concurrency groups which make it so that running actions auto-kills any existing runs for a given PR

rob-luke: Ah! I had been wondering what this did in the MNE-Python repo, thanks for explaining, that's a great feature.

doc/api.rst
```diff
@@ -156,47 +156,12 @@ Individual (first) level analysis functions.
...
-   RegressionResults.scatter
-   RegressionResults.plot_topo
-   RegressionResults.surface_projection
-   RegressionResults.scatter
-   RegressionResults.save
-   
-   ContrastResults.to_dataframe
-   ContrastResults.plot_topo
-   ContrastResults.scatter
-   ContrastResults.save
```
rob-luke, line 198: I really want to have these methods listed on the API page. Many users will be new to programming and won't understand about classes etc. So I fear that they will come to the API page, hit ctrl-f and search for "region", it will return nothing, and they will leave. Is there any way this is possible?

If not possible (which would make me sad), I would like to list them in plain text and point the user to look in the sub-api pages, so that the terms pop up in a ctrl-f search.

Thoughts?

rob-luke (changes requested):
First off, THANK YOU @larsoner. You've touched many things that I have been pulling my hair out about. I really want to get numpydoc working again, I appreciate your help.

doc/api.rst
larsoner: You should look at how scipy does it. It's not as easy as what MNE does but it does what you want I think. Personally I think the search box plus example linking is good enough for discoverability that this probably isn't really an issue for many folks but that's just speculation on my part

larsoner: Actually I'm not even sure if SciPy's allows listing them separately, even there I think they're only listed on the class's page.

One option would be too keep this but use a noindex template. I can try implementing that tomorrow...

larsoner:
This seems to do the right thing:

https://1668-254026779-gh.circle-artifacts.com/0/dev/pull/355/api.html#statistics

rob-luke:
This is just perfect. I love it. Thanks @larsoner

PR #360, opened 2021-08-18
Make figure example more generalisable

rob-luke:


examples/general/plot_16_waveform_group.py
```diff
@@ -293,7 +293,7 @@ def individual_analysis(bids_path):
                                  show=False, colors=[color], legend=False,
                                  ylim=lims, ci=0.95, show_sensors=cidx == 2)
             axes[0, cidx].set_title(f"{evoked}")
-            axes[1, cidx].set_title("") 
+            axes[ridx, cidx].set_title("")
```
larsoner, line 296: Clearer (and correct for the one-row case) as:
```
            axes[ridx, cidx].set_title("")
        axes[0, cidx].set_title(f"{evoked}")
```

rob-luke:
Thanks @larsoner!

PR #364, opened 2021-08-30
Specify h5py string type to match Homer

rob-luke:


rob-luke:
Just got word from the Homer camp that this was an issue with their reader, not with our writer. So closing this.

PR #372, opened 2021-09-08
Remove mayavi from requirements

rob-luke:
Follows #365 as the maintenance upkeep is too much for me.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -27,7 +27,6 @@ dependencies:
 - pyvistaqt>=0.4
 - qdarkstyle
 - darkdetect
-- PySurfer
 - dipy
 - nibabel
 - python-picard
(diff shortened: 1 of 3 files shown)


rob-luke:
Breaks more than it fixes. Cancelling.

agramfort:
interesting attempt. What does it break? you should also remove pysurfer that requires mayavi. We will want to do this soon

rob-luke:
Only circle ci broke. GH actions worked. As this is useful to others too I will try this again then. Starting with your suggestion. Then maybe this will help when you get around to it too.

rob-luke:
Thanks @agramfort now it works. I checked the CircleCI docs and the renders still look good, so nothing seems broken. Cheers.

PR #385, opened 2021-09-16
Update citation reference

rob-luke:
Really I was just trying to figure out what the issue with the failing CI was.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ scikit-learn
 nibabel
 numba
 h5py
-pandas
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -11,7 +11,7 @@ scikit-learn
 nibabel
 numba
 h5py
-pandas
+pandas==1.3.2
```
rob-luke, line 14: FYI @larsoner
There seems to have been some change in the latest pandas. Just letting you know incase this affects you somewhere

PR #388, opened 2021-09-20
hackish workaround for changes to downloading

drammock:
@rob-luke this is not great and should really wait for https://github.com/mne-tools/mne-python/pull/9763 but maybe this is enough of a stopgap for now?

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -35,6 +35,7 @@ dependencies:
 - ipywidgets
 - lxml
 - pytables
+- pooch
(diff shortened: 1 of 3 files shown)


rob-luke:
Thanks @drammock!! Stopgap is good enough for me for now. And we can follow up when the MNE-Python PR is merged. Thanks for doing this!

rob-luke:
Thanks @drammock

PR #394, opened 2021-09-28
Use MNE-Python fetch_dataset function

rob-luke:
#### Reference issue
Following https://github.com/mne-tools/mne-python/pull/9763 and #388 this correctly addresses the downloading of datasets in MNE-NIRS

Use the new `fetch_dataset` function from MNE-Python

### DIFF
--- a/mne_nirs/datasets/fnirs_motor_group/fnirs_motor_group.py
+++ b/mne_nirs/datasets/fnirs_motor_group/fnirs_motor_group.py
@@ -30,11 +30,10 @@ def data_path(path=None, force_update=False, update_path=True, download=True,
         config_key='MNE_DATASETS_FNIRSMOTORGROUP_PATH',
     )
 
-    dpath = fetch_dataset(dataset_params, processor=pooch.Unzip(
-        extract_dir="./fNIRS-motor-group"),
-                         path=path,
-                         force_update=force_update, update_path=update_path,
-                         download=download)
+    dpath = fetch_dataset(dataset_params, path=path, force_update=force_update,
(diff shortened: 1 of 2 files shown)


rob-luke:
@adam2392 could you lend me your experience with pooch and the new fetch_dataset to help me here? I am trying to create the directory structure I had before, which was ...

```text
/Users/rluke/mne_data/fNIRS-motor-group
├── CITATION.cff
├── README.md
├── dataset_description.json
├── participants.json
├── participants.tsv
├── sub-01
│   ├── nirs
│   │   ├── sub-01_coordsystem.json

```

But with my attempt at using the new fetch_dataset I get an extra subfolder which I don't want...


```text
/Users/rluke/mne_data/fNIRS-motor-group
└── BIDS-NIRS-Tapping-master
    ├── CITATION.cff
    ├── README.md
    ├── dataset_description.json
    ├── participants.json
    ├── participants.tsv
    ├── sub-01
    │   ├── nirs
    │   │   ├── sub-01_coordsystem.json
```

Is there a simple trick I am missing to download the data from GitHub and extract it in to ~/mne_data without adding an extra layer of directories? Any tips appreciated

adam2392:
The reason is due to these processors: https://www.fatiando.org/pooch/latest/api/generated/pooch.Untar.html

Your issue is that your desired folder name is not the same as the folder name that is extracted for your dataset. Thus you would like to extract contents of the zipped folder (not the folder itself) into a new folder name. 

Ideally, one could just specify `members=<your zipped folder>` in the pooch Processor, but seems like it can't be done yet.

PR #403, opened 2021-10-25
Automatically sanitse annotation names that arent compatible with nilearn

rob-luke:
#### Whats the problem?

Nilearn recently changed the requirements for design matrix column names in https://github.com/nilearn/nilearn/pull/3025 which disallowed strings containing only numbers and strings with `/` as column names. 

Getting rid of the numbers is no problem, I prefer meaningful annotation names anyway.

However, I think its common in MNE to use names like `tapping/left` or `response/right`. See https://mne.tools/dev/auto_tutorials/epochs/40_autogenerate_metadata.html?highlight=hed#keeping-only-the-first-events-of-a-group

But nilearn now requires that the column names pass the following test:

```python
'response/right'.isidentifier()
# False

'response_right'.isidentifier()
# True
```

See lines: https://github.com/nilearn/nilearn/blob/ab9c68e502c4b92695cd658f5839a8207fd5e07e/nilearn/glm/first_level/hemodynamic_models.py#L428-L436

#### Whats the fix?

This PR checks that our design matrix column names are compatible with nilearn, and if they are not then prepends a `t_` to force them to be valid.

#### The problem

This PR also changes `/` to `_` in each design matrix column name, but this is not then harmonious with the rest of MNE. Further, it means that in some places of my script I use `tapping/left` and other places I will need to use `tapping_left`. This solution doesn't seem great, what is a better alternative?

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -31,6 +31,11 @@ Enhancements
 * Add :func:`mne_nirs.io.fold_channel_specificity`. By `Robert Luke`_.
 
 
+Fixes
+
+* Prepend a character to event names that are number only to comply with nilearn criteria for design matrix column names. By `Robert Luke`_.
+
+
 v0.1.1
 ------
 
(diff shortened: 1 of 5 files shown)


rob-luke:
We now know that we can not use the `/` symbol in the design matrix. So I will try and write some sort of sanitisation approach as suggested by larsoner and will use `___` to equal `/`.

PR #417, opened 2021-11-30
ENH: Speed up CircleCI

larsoner:
1. Fix the `numpydoc` build error about plot_gallery being a str
2. Speed up CircleCI by using a machine image
3. Get rid of a warning (hopefully) that we see at the start of the doc build

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,6 +18,7 @@ jobs:
             name: Set BASH_ENV
             command: |
               set -e
+              python3 -m venv ~/python_env --upgrade-deps
               echo "set -e" >> $BASH_ENV
               echo "export DISPLAY=:99" >> $BASH_ENV
               echo "export OPENBLAS_NUM_THREADS=4" >> $BASH_ENV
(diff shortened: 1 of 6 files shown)


rob-luke:
Amazing! I've been hating those errors and warnings!

larsoner:
@rob-luke it looks like CircleCI has been running for >35 minutes. We're going to start paying for CircleCI time soon, so it would be good to cut this down if possible. It looks like PRs are currently doing all builds using sphinx-multiversion (right?), would it be an option for PRs to just build the current PR version of the docs rather than all versions? You could adopt a syntax like we have for MNE-Python that could even cause it to build all versions rather than just the current dev one for example.

rob-luke:
> would it be an option for PRs to just build the current PR version of the docs rather than all versions?

Yes. This is how the github doc building works, it only builds the PR branch. This was handled in the github doc building using...

https://github.com/mne-tools/mne-nirs/blob/8b9e7b854b6308ed4067e778c2fad6a277fa1f8b/.github/workflows/SphinxBuild_PR.yml#L23

and maybe...

https://github.com/mne-tools/mne-nirs/blob/8b9e7b854b6308ed4067e778c2fad6a277fa1f8b/.github/workflows/SphinxBuild_PR.yml#L59-L63

larsoner:
Feel free to merge this if you're happy @rob-luke , we can deal with CircleCI speed in https://github.com/mne-tools/mne-nirs/pull/418

rob-luke:
Thanks @larsoner

PR #418, opened 2021-11-30
MRG, MAINT: Faster CircleCI

larsoner:
An alternative solution to the doc-build issue @rob-luke -- just use `sphinx` directly. Not sure if the SMV_TAG_WHITELIST approach is better... but it's nice to be able to do `make html-single` locally, too, I think!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -126,7 +126,7 @@ jobs:
               git status
               git tag
               git branch
-              make html;
+              make html-single;
 
         # Save the outputs
         - store_artifacts:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,20 @@ jobs:
       steps:
         - checkout
         - run:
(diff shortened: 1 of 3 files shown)


larsoner:
15m total runtime, not too bad for building all the examples

rob-luke:
My only hesitation is that now there is two methods for me to maintain, the multisphinx and plain sphinx, but thats manageable (with your help :wink: ).

Alternatively, it looks like github actions is going to work now with the 3d figures. So we could just delete circleci altogether?

larsoner:
The big advantage of circle is that you can immediately view the artifact to see how it looks

rob-luke:
> The big advantage of circle is that you can immediately view the artifact to see how it looks

Agreed. Lets proceed and merge this then. Thanks!

larsoner:
... And you are already maintaining a plain sphinx build with the numpydoc target on actions :)

PR #421, opened 2021-12-08
Validate SNIRF file writer

rob-luke:
Validate that the SNIRF files we write are valid using the official validator: https://github.com/BUNPC/pysnirf2 thanks to @sstucker

Our files **did** pass the old validator https://github.com/andyzjc/snirf_validator

Currently our files **are not** passing the new validator, so we may need to modify the MNE SNIRF writer.


sstucker:
I would wait to let a few kinks get worked out of the validation code.

One thing that I am confident is a good goalpost is proper string formatting via h5py which is required by the validator (or else a warning will be raised)

```python
_varlen_str_type = h5py.string_dtype(encoding='ascii', length=None)  # Length=None creates HDF5 variable length string

def _create_dataset_string(file: h5py.File, name: str, data: str) -> h5py.Dataset:
    return file.create_dataset(name, dtype=_varlen_str_type, data=str(data))
```

This consistently exports a _variable length string_/ascii null-terminated string. If a mixture of variable length and fixed length strings are used the parsing logic becomes more complicated so we are trying to move towards just one: https://github.com/fNIRS/snirf/issues/72

Another thing that happens a lot because is values being saved as arrays of 1 element and 1 dimension. This will cause the validator to mark your Dataset with INVALID_DATASET_SHAPE.

rob-luke:
Thanks @sstucker
It seems our files now pass the validator, so I will merge this additional test to ensure we dont break SNIRF comparability in the future.

PR #422, opened 2021-12-13
Add custom channel weighting functionality in GLM region of interest

rob-luke:
#### Reference issue

Currently the region of interest calculation has two weighting options:
1. Weight each channel equally (`weighted=False`)
1. Weight each channel by the inverse of the standard error fo the GLM fit (`weighted=True`)

#### Problem / Use Case

Each fNIRS channel is located at a different position on the scalp and has different sensitivity to brain regions of interest. A user may wish to weight the channels in their region of interest according to the sensitivity to desired brain structure. I.e. weight channels with better sensitivity to brain structure higher than channels with low sensitivity.

Adds a third option when calculation region of interest results from individual channels

3. Custom channel weighting. Allows the user to manually specify the relative weighting of each channel in the ROI computation.

The user can specify a dictionary in the same format as the `group_by` parameter. The channels are then weighted according to the custom values. The values are normalised to one, so that the scaling is not altered (this was already implemented for the SE option).

#### Additional information
This is useful in conjunction with other software that provides channel by channel sensitivity to desired brain regions such as fOLD, AtlasViewer, etc

Also fixed a bug when `weighted=False` was used. This was not being normalised to one, so results were scaled. I never used this, so never noticed it before, just caught it when making the tests.

#### TODO

- [x] More tests
- [x] Example in docs

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -22,6 +22,12 @@ Enhancements
 
 * Add ability to provide custom channel weighting in :meth:`mne_nirs.statistics.RegressionResults.to_dataframe_region_of_interest` computation. By `Robert Luke`_.
 
+
+Fixes
+
+* Fix bug when using no weighting in :meth:`mne_nirs.statistics.RegressionResults.to_dataframe_region_of_interest`. By `Robert Luke`_.
+
+
 v0.1.2
 ------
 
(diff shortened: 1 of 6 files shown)


rob-luke:
@larsoner do you have any feedback on this API change?

rob-luke:
Tutorial render: https://2059-254026779-gh.circle-artifacts.com/0/dev/auto_examples/general/plot_70_visualise_brain.html#apply-anatomically-informed-weighting-to-region-of-interest-analysis

PR #427, opened 2021-12-31
Test if CIs are running correctly

rob-luke:
Looks like the CIs are broken. I'll try and align with the latest files from MNE-Python and see if that fixes things.


rob-luke:
Need to align MNE-NIRS requirements and CI code with MNE-Python. But I won't have time to do this before Feb.

rob-luke:
See #429

PR #426, opened 2021-12-31
Pandas version unpin

FuriouslyCurious:
#### Reference issue
Fixes #425

#### What does this implement/fix?
Removed pandas version pinning

#### Additional information
I recommend devs take a look at pyqt5 dependencies as well.

PR #431, opened 2022-01-21
Unpin Pandas

FuriouslyCurious:
Resubmitting #426 after rebasing...

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ scikit-learn
 nibabel
 numba
 h5py
-pandas==1.3.2
+pandas>=1.3.2
 numexpr
 jupyter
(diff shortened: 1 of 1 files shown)


rob-luke:
Ok, this has thrown some errors. To check if it's really caused by the pinning can you repin, check if the tests go green, then unpin again and check they go red. I'd do it myself, but you must have deselected the option to allow maintainers to edit the branch.

But I have a feeling this might be a real bug/issue and thats why the version was pinned originally. But lets tackle that if you confirm the errors are related to the version.

rob-luke:
Actually I am seeing a similar error in #432 so maybe statsmodels has released a new version that breaks something. Ill check

PR #432, opened 2022-01-24
Extra filtering example

rob-luke:


tools/github_actions_dependencies.sh
```diff
@@ -14,8 +14,8 @@ else
 	pip install --progress-bar off --upgrade --pre --only-binary ":all:" python-dateutil pytz joblib threadpoolctl six
 	echo "PyQt5"
 	pip install --progress-bar off --upgrade --pre --only-binary ":all:" --no-deps --extra-index-url https://www.riverbankcomputing.com/pypi/simple PyQt5 PyQt5-sip PyQt5-Qt5
-	echo "NumPy/pandas etc."
-	pip install --progress-bar off --upgrade --pre --only-binary ":all:" --no-deps -i "https://pypi.anaconda.org/scipy-wheels-nightly/simple" numpy pandas "scikit-learn>=0.24.2" dipy
+	echo "NumPy etc."
+	pip install --progress-bar off --upgrade --pre --only-binary ":all:" --no-deps -i "https://pypi.anaconda.org/scipy-wheels-nightly/simple" numpy "scikit-learn>=0.24.2" dipy
```
larsoner, line 18: It's not great if you aren't testing `--pre` pandas, because it means the next time they cut a release user code will probably break. (You could require a specific version of pandas in `setup.py`, but then you could break someones' environment, which is worse than breaking one use case.)

Do you need help looking into the pandas issue here?

rob-luke: > Do you need help looking into the pandas issue here?

Yes please @larsoner I havent started digging yet, I was just focused on getting the CIs green first. but seems I may have made more issues than I solved. Thanks

PR #442, opened 2022-02-04
ENH: plot_3d_montage

### LINKED ISSUE #440: ENH: Montage plots for fNIRS
Not sure if this should go in MNE-Python, MNE-NIRS, or neither (?), but I've repeatedly wanted to show (in papers, and just to make sure I'm doing things correctly) plots of fNIRS electrodes labeled with the 10-20 coordinates they have when they are being positioned on the subject. However:

1. fNIRS channels have specific namings related to their source-detector pairings and wavelengths, so `rename_channels` is not a good option
2. The channels we actually record are a combination of a source-detector pair, like the yellow dots [here](https://mne.tools/dev/auto_tutorials/preprocessing/70_fnirs_processing.html#viewing-location-of-sensors-over-brain-surface):

    

3. However, the optodes that are placed on the heads are actually sources and detectors, i.e., the red and dark red/brown above.

What I've found really useful for sanity checking is a multi-view plot with ideal 10-20 locations, plus the resulting fNIRS channel numbers (starting from 1; here for a Hitachi system):

montage

For MNE-Python, it might be useful to add text labels to channels in `plot_alignment`, though I'm not sure how nicely these could be made to interact with 3D view manipulations. (My code currently uses the given 3D-to-2D projection to add 2D text actors. Proper 3D interaction would be tougher, since we'd need to place the text in the 3D scene -- probably with some offset "outward" so that it doesn't overlap with the channel sphere -- and make sure it draws nicely with translucent geometry and such).

So I think overall I'm inclined to just implement some sort of `mne_nirs.viz.plot_3d_montage` or something. If folks agree about that scope, then I'll transfer this issue over to the MNE-NIRS repo and hash out an API with @rob-luke.

larsoner:
Closes #440

Code is getting there. If MNE-Python's montage functions are used, autodetection works:
```
    view_map = {
        'left-lat': np.arange(1, 13),
        'right-lat': np.arange(13, 25),
    }
    fig = mne_nirs.visualisation.plot_3d_montage(
        use['h'].info, view_map=view_map, subjects_dir=subjects_dir)
```
montage

NIRx doesn't use the same positions as the builtin MNE-Python montage, so we need to do more for this case:
```
    view_map = {
        'left-lat': np.r_[np.arange(1, 13),
                          np.arange(58, 74),
                          np.arange(75, 78)],
        'caudal': np.r_[42,
                        np.arange(45, 58),
                        74,
                        np.arange(78, 85)],
        'right-lat': np.r_[np.arange(13, 42), 43, 44],
    }
    fig = mne_nirs.visualisation.plot_3d_montage(
        info, view_map=view_map, subjects_dir=subjects_dir)
```

montage

The [modified example](https://2145-254026779-gh.circle-artifacts.com/0/dev/auto_examples/general/plot_70_visualise_brain.html#plot-sensor-channel-numbers) has the same problem:
```
view_map = {
    'left-lat': np.r_[np.arange(1, 27), 28],
    'caudal': np.r_[27, np.arange(43, 53)],
    'right-lat': np.r_[np.arange(29, 43), 44],
}

fig = mne_nirs.visualisation.plot_3d_montage(
    raw.info, view_map=view_map, subjects_dir=subjects_dir)
```
Screenshot from 2022-02-04 15-40-10

# Todo

- [ ] ~~Fix NIRx 10-20 detection~~
- [x] Actually describe the `view_map` option (hemi-brainview keys)
- [x] Add tests

## Option 1

@rob-luke one option would be to make a function for NIRx that converts their 10-20 positions to MNE-Python's. Do they have standard 10-20 positions that then populate `info`? If so, we could come up with a lookup table, then people can do:
```
mne_nirs.something.convert_nirx_1020(raw.info, copy=False)
```
to update their channel positions. Users can even avoid making it permanent just by doing in their code before plotting:
```
plot_info = mne_nirs.something.convert_nirx_1020(raw.info)  # default is copy=True
mne_nirs.visualisation.plot_3d_montage(plot_info, ...)
```

## Option 2

Another clean way to go would be, instead of `'auto'`, the default value could be `src_det_names='standard_1020'` to give a montage to load. Then in MNE-Python we can add a `'nirx'` named montage, so then the call here would be `src_det_names='nirx'` or so. Or we just make `nirx_montage = mne_nirs.something.get_montage('nirx')` that you can then pass directly as `src_det_names=nirx_montage` to tell it which positions to look at to get the names.

I think my vote is for Option 2, because it nicely generalizes. You could imagine generalizing to `get_montage('flow')` or so. And then `src_det_names=<str>` could by convenience call the mne_nirs get_montage first, and if it doesn't exist, try loading the MNE-Python get_montage (like standard_1020)... and we could maybe even make `auto` detect the acq system and triage to the correct one (standard_1020 for Hitachi, nirx for nirx, flow for kernel flow, etc.).

### DIFF
--- a/mne_nirs/visualisation/_plot_3d_montage.py
+++ b/mne_nirs/visualisation/_plot_3d_montage.py
@@ -116,7 +116,7 @@ def plot_3d_montage(info, view_map, *, src_det_names='auto',
     brain = Brain(
         subject, 'both', 'pial', views=['lat'] * len(views),
         size=size, background='w', units='m',
-        view_layout='horizontal')
+        view_layout='horizontal', subjects_dir=subjects_dir)
     with _safe_brain_close(brain):
         brain.add_head(dense=False, alpha=0.1)
         brain.add_sensors(
--- a/mne_nirs/visualisation/_plot_3d_montage.py
+++ b/mne_nirs/visualisation/_plot_3d_montage.py
@@ -17,7 +17,7 @@ from mne.viz import Brain
 
 @verbose
 def plot_3d_montage(info, view_map, *, src_det_names='auto',
-                    subject='fsaverage', trans='fsaverage',
+                    subject='fsaverage', trans='fsaverage', surface='pial',
                     subjects_dir=None, verbose=None):
     """
     Plot a 3D sensor montage.
@@ -48,6 +48,8 @@ def plot_3d_montage(info, view_map, *, src_det_names='auto',
         The subject.
     trans : str | Transform
         The subjects head<->MRI transform.
+    surface : str
+        The FreeSurfer surface name (e.g., 'pial', 'white').
     subjects_dir : str
         The subjects directory.
     %(verbose)s
@@ -100,7 +102,8 @@ def plot_3d_montage(info, view_map, *, src_det_names='auto',
             logger.info('Source-detector names automatically mapped to 10-20 '
                         'locations')
 
-    trans = _get_trans('fsaverage', 'head', 'mri')[0]
+    head_mri_n = _get_trans(trans, 'head', 'mri')[0]
+    del trans
     views = list()
     for key, num in view_map.items():
         _validate_type(key, str, f'view_map key {repr(key)}')
@@ -114,13 +117,13 @@ def plot_3d_montage(info, view_map, *, src_det_names='auto',
     del view_map
     size = (400 * len(views), 400)
     brain = Brain(
-        subject, 'both', 'pial', views=['lat'] * len(views),
+        subject, 'both', surface, views=['lat'] * len(views),
         size=size, background='w', units='m',
         view_layout='horizontal', subjects_dir=subjects_dir)
     with _safe_brain_close(brain):
         brain.add_head(dense=False, alpha=0.1)
         brain.add_sensors(
-            info, trans='fsaverage',
+            info, trans=head_mri_t,
             fnirs=['channels', 'pairs', 'sources', 'detectors'])
         add_text_kwargs = dict()
         if 'render' in _get_args(brain.plotter.add_text):
@@ -149,8 +152,7 @@ def plot_3d_montage(info, view_map, *, src_det_names='auto',
                     if name in plotted:
                         continue
                     plotted.add(name)
-                    # name = rev_dict[use_]  # XXX ADD THIS
-                    ch_pos = apply_trans(trans, ch_pos)
+                    ch_pos = apply_trans(head_mri_t, ch_pos)
                     vp.SetWorldPoint(np.r_[ch_pos, 1.])
                     vp.WorldToDisplay()
                     ch_pos = (np.array(vp.GetDisplayPoint()[:2]) -
--- a/mne_nirs/visualisation/_plot_3d_montage.py
+++ b/mne_nirs/visualisation/_plot_3d_montage.py
@@ -154,7 +154,7 @@ def plot_3d_montage(info, view_map, *, src_det_names='auto',
                     vp.SetWorldPoint(np.r_[ch_pos, 1.])
                     vp.WorldToDisplay()
                     ch_pos = (np.array(vp.GetDisplayPoint()[:2]) -
-                            np.array(vp.GetOrigin()))
+                              np.array(vp.GetOrigin()))
 
                     actor = brain.plotter.add_text(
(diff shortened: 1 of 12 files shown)


mne_nirs/visualisation/_plot_3d_montage.py
```diff
@@ -0,0 +1,149 @@
...
+    Plot a 3D sensor montage.
+
+    Parameters
+    ----------
+    info : instance of Info
+        Measurement info.
+    view_map : dict
+        Dict of view (key) to channel-pair-numbers (value) to use when
+        plotting. Note that, because these get plotted as 1-based channel
+        *numbers*, the values should be 1-based rather than 0-based.
```
larsoner, line 30: I also need to actually describe the options here (hemi-brainview keys)

rob-luke:
This looks really cool @larsoner 

> @rob-luke one option would be to make a function for NIRx that converts their 10-20 positions to MNE-Python's. Do they have standard 10-20 positions that then populate info? 

You can place your optode positions anywhere in a NIRX system, you are not restricted to 10-20 style locations. When i've built my montages I just click as close to the 10-20 positions as I can.  So there is not a direct mapping, does this make sense?

But in general, the NIRS community haven't used these standard positions as much as they could (and I have received much critisim in manuscript reviews for describing locations using "eeg" language). But I hope in the future the fNIRS community adopts standard locations more.

As most devices dont restrict users to 10-20 positions, I dont know if your option 2 will work. Instead of specifying positions using 10-20 naming, all the fNIRS data files I have used specify the locations using 3D coordinates. But I guess when using `auto` we could try and detect a standard position name, and if the distance/error is too great then just revert to the uninformative naming of `S1` etc? I dont knot if I am being helpful here or not.

larsoner:
> When i've built my montages I just click as close to the 10-20 positions as I can. So there is not a direct mapping, does this make sense?

Ahh so even if in experiment 1 you placed an electrode on the participant and clicked in their GUI as close to Cz as you can, these two could (probably will) end up with different X/Y/Z coordinates? Then in this case I think the solution here is probably all we want -- if the electrode positions match the standard MNE 10-20, use 10-20 names, otherwise rely on the user to either 1) `set_montage` using the MNE standard 10-20 locations if they want to standardize or 2) pass a mapping dict directly to say what S1, S2, ..., D1, D2, ... map to.

I'll move on to writing some tests and improving the docstring, then I think that's probably sufficient for this first PR.

rob-luke:
> Ahh so even if in experiment 1 you placed an electrode on the participant and clicked in their GUI as close to Cz as you can, these two could (probably will) end up with different X/Y/Z coordinates? 

Correct. But I do this only once per project/montage, then apply the same montage to all measurements within a single experiment.

> Any objection to adding an alias mne_nirs.viz that's the same as mne_nirs.visualisation? It bypasses the british/american English issue

I would have gone with vis 😉 But I agree we should go with viz for consistency with MNE etc

Ping me when its ready for review

larsoner:
Okay @rob-luke this is ready for review/merge from my end

mne_nirs/visualisation/_plot_3d_montage.py
```diff
@@ -0,0 +1,178 @@
...
+        The keys are of the form:
+
+        ``'{side}-{view}'``
+            For views like ``'left-lat'`` or ``'right-frontal'`` where the side
+            matters.
+        ``'{view}'``
+            For views like ``'caudal'`` that are along the midline.
+
+        See :meth:`mne.viz.Brain.show_view` for ``view`` options.
+    src_det_names : None | dict | str
```
rob-luke, line 42: Could you please add another example to the docs that uses the explicit setting of names via dict? Can be just directly below the existing example.

rob-luke (changes requested):
Thanks @larsoner

larsoner:
@rob-luke done!

https://2158-254026779-gh.circle-artifacts.com/0/dev/generated/mne_nirs.visualisation.plot_3d_montage.html#mne_nirs.visualisation.plot_3d_montage

rob-luke:
This is so great. Thanks @larsoner

rob-luke:
Just added a changelog entry. Merging now, thanks.

larsoner:
Feel free to make it better. In particular, the short channel plotting is a bit ugly. Might want some way to disable short channel labeling, or maybe show it differently somehow...

rob-luke:
Good ideas!

PR #443, opened 2022-02-09
Check SNIRF files are valid

rob-luke:
Bump to kick start the CI and test our SNIRF files against the newly released validator from @sstucker at https://github.com/BUNPC/pysnirf2/pull/27

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -19,7 +19,7 @@ MNE-NIRS: Near-Infrared Spectroscopy Analysis
 .. image:: https://badge.fury.io/py/mne-nirs.svg
     :target: https://badge.fury.io/py/mne-nirs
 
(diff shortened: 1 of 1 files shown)


rob-luke:
Looks like our files are valid according to the new validator. Nothing to do here then. Closing.

sstucker:
I will make sure to alert you to a big fix in the validator if you'd like. This one was minor--fixing a bug that caused a crash on certain invalid files.

rob-luke:
If you remember, but no stress, I've got a watch on your repo 😉 and will try and kick off my CI manually each time you push an update. Thanks for the great package.

larsoner:
> will try and kick off my CI manually each time you push an update. Thanks for the great package.

Depending on the update frequency (daily? weekly?) I would set up an Azure cron job that runs and notifies you by email on failure. I set one up recently @rob-luke so I can help you do it if you want

PR #446, opened 2022-02-18
MAINT: Deal with Path->str dep

larsoner:
Need to fix this so I can actually see if there are errors in https://github.com/mne-tools/mne-nirs/pull/445


larsoner:
@rob-luke I'll merge this if CIs come back green, but feel free to review and if you think an alternative approach is better, I'm happy to make another PR

PR #450, opened 2022-02-28
MAINT: Make work with latest stable

### LINKED ISSUE #448: MAINT: Testing policy
For maintainability, it's usually good to maintain compatibility with at least the current stable version of MNE-Python. @rob-luke okay to adopt this policy, and formalize it via CIs testing against MNE-Python `main` as well as the latest `pip`/`conda`-installable version?

larsoner:
Closes #448

Takes the easy way out of just skipping tests that fail on 0.24. Most of these skips can be removed once MNE-Python 1.0 comes out. But in the meantime, it sets us up to regularly test stable + main.

Adds an azure-pipelines.yml that will:
1. Run on all PRs with MNE-Python `main`
2. Tests on windows (new)
3. Runs nightly on all platforms

Once this is merged, I'll use the Azure interface to email me and you @rob-luke when the nightly build fails. I got it working for MNE-KIT-GUI at some point, it should be possible to do it here, too.


rob-luke:
Looks good @larsoner, merge when you want.

larsoner:
Finally green! In it goes

PR #452, opened 2022-03-01
ENH: Allow setting ch_names

larsoner:
It's nice to be able to remap the `ch_names` that get plotted. I'm not convinced this is the best API. Maybe instead we should have:
```
ch_names : str | dict
    Channel names to plot. Can be ``'numeric'`` (default) to use ``['1', '2', ...]``,
    ``'short'`` to use ``info['ch_names'][::2]`` (e.g., ``['S1_D2', 'S2_D1', ...]``),
    or a dict to provide a mapping from ``info['ch_names'][::2]`` (keys) to the
    desired names.
```
@rob-luke WDYT?

### DIFF
--- a/mne_nirs/visualisation/_plot_3d_montage.py
+++ b/mne_nirs/visualisation/_plot_3d_montage.py
@@ -49,7 +49,7 @@ def plot_3d_montage(info, view_map, *, src_det_names='auto',
         for example::
 
             src_det_names=dict(S1='Fz', D1='FCz', ...)
-    ch_names : str | dict
+    ch_names : str | dict | None
         If ``'numbered'`` (default), use ``['1', '2', ...]`` for the channel
         names, or ``None`` to use ``['S1_D2', 'S2_D1', ...]``. Can also be a
         dict to provide a mapping from the ``'S1_D2'``-style names (keys) to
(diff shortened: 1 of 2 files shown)


rob-luke:
I love it! I can see myself setting `src_det_names=None` and then using a dict to map S2_D3 = "Planum Temporale" and S5_D2 = "Inferior Frontal Gyrus" etc.

Would this allow new line characters and/or setting S7_D3 = "", i.e. removing channel names for certain channels?

larsoner:
Removing definitely works. Newlines I'll have to check

rob-luke:
> Newlines I'll have to check

If not, no need to add it here, just make a new issue and we can tackle it later.

larsoner:
Newlines work -- I used `src_det_names=defaultdict(lambda: 'a\nb')`:

Screenshot from 2022-03-01 19-13-13

So do you think the API should be as described at the top of the issue, or as it's implemented in the code?

rob-luke:
> So do you think the API should be as described at the top of the issue, or as it's implemented in the code?

I prefer as described at the top of the issue. I think that `ch_names ="short"` is much more approachable to a new user than lambdas or `dict(zip(str(ii), name.split()[0]) for ii, name in enumerate(raw.ch_names[::2], 1)))`

larsoner:
Yes but shouldn't we teach them this beautiful one-liner :)

I'll push a fix tomorrow (EST)

larsoner:
Ready for review/merge from my end @rob-luke

rob-luke:
Wonderful, thanks @larsoner

PR #455, opened 2022-03-02
MAINT: Add nilearn dep

### LINKED ISSUE #454: nilearn dependency missing from setup.py
Simply trying to import the package after installation will fail unless `nilearn` is installed. Hence, `nilearn` should be listed as a dependency in `setup.py`.

larsoner:
@rob-luke 0.9 was released about a month ago (Jan 28). Should we require 0.8 instead?

Closes #454


rob-luke:
0.9 changed the way that annotation names were treated and validated etc. So code that works with 0.8 doesn't necessarily work with 0.9+. 

So I think we should support 0.9 and above only.

rob-luke:
Seems the CIs aren't happy though :(

larsoner:
Failures were just not having the fOLD files, fixed. I'll go ahead and merge. We should backport this and release a 0.2.1. I'll actually wait to do that until the doc buliding is fixed, hopefully tomorrow! Then we can see deployment of 0.2.1 docs and 0.3.dev0 docs, hopefully.

PR #458, opened 2022-03-05
MAINT: Update dataset hash

larsoner:
New commit changed the hash of BIDS-NIRS-tapping

https://app.circleci.com/pipelines/github/mne-tools/mne-nirs/2245/workflows/eadc14d3-a7f8-4dcf-9adc-97b54a0d224a/jobs/2256

This also changes to using a fixed commit number (the new one) so that this won't happen again. If the new one doesn't work, I'll switch back. When the dataset is updated, a PR should be made here to update instead. (And the CircleCI cache number probably needs to be incremented.)

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -245,6 +245,12 @@ else:
     report_scraper = mne.report._ReportScraper()
     scrapers += (report_scraper,)
     del backend
+try:
+    import mne_qt_browser
+    if mne.viz.get_browser_backend() == 'qt':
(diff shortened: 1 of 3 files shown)


larsoner:
Argh we can't use a hash because then the archive has a subfolder `BIDS-NIRS-tapping-<hash>`. So this just bumps the hash and adds mne-qt-browser as the raw data browser, since that's preferred on MNE 1.0 now. Will merge once CIs are happy, then hopefully I can deal with any deployment issues!

PR #469, opened 2022-07-01
Why are CI's failing?

rob-luke:
Investigation 👀 

* There was an upstream change to how the theme handles different versions
* Dev is failing due to mne handling of scipy changes (no action here yet)


rob-luke:
This doesn't fix the dev issues, but gets us closer to all green. So in it goes

Issue #484, opened 2022-09-13
Stable documentation not building


rob-luke:
@larsoner I released a new version of MNE-NIRS and followed the instructions in https://github.com/mne-tools/mne-nirs/blob/main/doc/release.md. Everything has gone smoothly, except a new stable docs was not deployed to https://github.com/mne-tools/mne-nirs/tree/gh-pages. Do you have any suggestions as to where I might have gone wrong?

larsoner:
Probably this

https://github.com/mne-tools/mne-nirs/blob/38d5e34f06674fbb6a8b01f5e9985ceae11ef751/.circleci/config.yml#L236

I'll push a quick fix

larsoner:
https://github.com/mne-tools/mne-nirs/commit/46c85d4ec179b37d321d9b56ebf789d8c676bd6b

larsoner:
Looks like it deployed but there are some bugs remaining

<img width="206" alt="Screen Shot 2022-09-13 at 9 46 08 AM" src="https://user-images.githubusercontent.com/2365790/189959704-b723939c-ede1-4b69-b51a-4a893f0f5647.png">

Feel free to fix when you get a chance @rob-luke

larsoner:
(I got that dropdown when I selected 0.3.0)

rob-luke:
Thanks @larsoner ill address the remaining bug(s) and update the instructions to include your fix above. Thanks mate

rob-luke:
Sorry to bug you again @larsoner 
It seems that the fix did not cause the stable docs to be built. Looking at https://github.com/mne-tools/mne-nirs/tree/gh-pages it shows that stable was last changed 6 months ago (dev was changed 3 days ago at last commit to main).
I guess this is why the dropdown is wrong. Could you take a look at why stable didn't get regenerated after the 0.3 release? I tried kicking the CI with a small push to the main/0.3 branch but that didnt work.

larsoner:
No problem! I always look at CircleCI directly:

https://app.circleci.com/pipelines/github/mne-tools/mne-nirs

You can see that your most recent builds are failing:

https://app.circleci.com/pipelines/github/mne-tools/mne-nirs/2416/workflows/c28e631e-8676-4053-97f4-f61a1e5ef936/jobs/2454

[...]

PR #485, opened 2022-09-17
MAINT: Work around zmq wheel bug

rob-luke:
Relevant to #484 
And based on https://github.com/mne-tools/mne-python/pull/11169/files

### DIFF
--- a/requirements_doc.txt
+++ b/requirements_doc.txt
@@ -23,3 +23,4 @@ snirf
 statsmodels
 dabest
 mne-qt-browser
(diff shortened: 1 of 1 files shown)


larsoner (approved):

larsoner:
Although looking at it now, I think they have managed to fix their wheels, so this is probably no longer necessary

rob-luke:
Instead I have a new colorbar issue to solve. haha it never ends 😆 !

PR #497, opened 2023-01-02
Fix SNIRF duration writing bug

rob-luke:
#### Reference issue
SNIRF file should write stims/annotations with the following format: `[starttime duration value]`.
As described here: https://github.com/fNIRS/snirf/blob/master/snirf_specification.md#nirsistimjdata

Modifes the `write_raw_snirf ` function to write stims according to the format above.
Added tests to ensure round trip read-write is lossless for stimulus parameters.

#### Additional information
This also requires a change to MNE here: https://github.com/mne-tools/mne-python/pull/11397

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -17,7 +17,7 @@ To install a specific version of the library you would run ``pip install mne-nir
 v0.6.dev0
 ---------
 
-* Fix bug in SNIRF writter that caused incorrect duration to be written to file. By `Robert Luke`_.
+* Fix bug in SNIRF writer that caused incorrect duration to be written to file. By `Robert Luke`_.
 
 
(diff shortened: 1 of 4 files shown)


rob-luke:
@larsoner could you please review. Specifically, I am unsure about my skipping of tests for pre 1.4 and if this is a bad idea?

larsoner:
Up to you if you want to skip. Another slightly more complete option would be to triage based on < 1.4 or >= 1.4 and check that the duration matches expectation, e.g. 5 on < 1.4 and the correct value(s) on >= 1.4. But this is probably overkill. Feel free to merge if you're happy with this version!

rob-luke:
Thanks mate, in it goes!

PR #499, opened 2023-01-12
Add AtlasViewer compatible snirf export

rob-luke:
AtlasViewer required template positions to be exported in the `landmarkLabels` field to perform alignment. This PR provides an option to export those landmarks with the snirf writer.

### DIFF
--- a/mne_nirs/io/snirf/_snirf.py
+++ b/mne_nirs/io/snirf/_snirf.py
@@ -17,7 +17,7 @@ from mne.channels import make_standard_montage
 SPEC_FORMAT_VERSION = '1.0'
 
 
-def write_raw_snirf(raw, fname, atlasviewer=False):
+def write_raw_snirf(raw, fname, add_montage=False):
     """Write continuous wave data to disk in SNIRF format.
 
     Parameters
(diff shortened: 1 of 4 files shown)


mne_nirs/io/snirf/_snirf.py
```diff
@@ -213,8 +239,24 @@ def _store_probe_landmarks(raw, probe):
...
+        ch_names = montage.ch_names
+        montage_locs = np.array(list(montage._get_ch_pos().values()))
+
+        # montage_locs = np.array([d['r'] for d in montage.dig])
+        head_mri_t, _ = _get_trans('fsaverage', 'mri', 'head')
+        locations_head = apply_trans(head_mri_t, montage_locs)
+
+        for template_idx in range(len(ch_names)):
+            digname.append(ch_names[template_idx])
+            diglocs = np.vstack((diglocs, locations_head[template_idx, :]))
```
larsoner, line 258: Simpler and vectorized?
```suggestion
        digname.extend(ch_names)
        diglocs = np.concatenate((diglocs, locations_head))
```

mne_nirs/io/snirf/_snirf.py
```diff
@@ -24,6 +26,9 @@ def write_raw_snirf(raw, fname):
         Data to write to file. Must contain only `fnirs_cw_amplitude` type.
     fname : str
         Path to the SNIRF data file.
+    atlasviewer : bool
+        Should additional landmarks be exported to
+        facilitate compatibility with AtlasViewer.
```
larsoner, line 31: To me a name like `add_montage` might be a better name since it's really adding standard montage channel positions. Then the description can say it facilitates use with atlasviewer. To me that part's secondary but maybe it's not...

rob-luke: great suggestion @larsoner you are correct

Issue #506, opened 2023-04-26
Fix all continuous integration errors


rob-luke:
#504 demonstrated that several of the CI's are currently failing.

- Ensure all CI's are working
- The CI system is a mirror of MNE-Python. So a good first step is to look at what changes have recently made to that infrastructure and mirroring the fixes here. E.g. look at their requirements.txt file and the associated PR, check recent changes to their .github directory, check their changes to .circleci directory etc
- Do not make fixes outside the scope of fixing the CI's (unless code changes are necessary to fix the CI). If you find any general code improvements or want to add features etc, then open a dedicated PR for that

PR #507, opened 2023-04-26
MAINT: Update dependencies and revitalize the CI pipeline

florin-pop:
This PR fixes https://github.com/mne-tools/mne-nirs/issues/506 by introducing the following changes:
 - have pip install packages directly from `git` instead of GitHub tarballs or zipballs
 - replace `footer_items` (which is deprecated) with `footer_start` in `doc/conf.py`
 - fixes a typo
 - install `libxcb-cursor0` in addition to the rest of libxcb dependencies

These fixes were mostly adaptations of changes introduced by @larsoner in https://github.com/mne-tools/mne-python/

---

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,7 +65,7 @@ jobs:
               echo "export OPENBLAS_NUM_THREADS=4" >> $BASH_ENV
               echo "export XDG_RUNTIME_DIR=/tmp/runtime-circleci" >> $BASH_ENV
               echo "export MNE_FULL_DATE=true" >> $BASH_ENV
-              echo "export MNE_3D_OPTION=false" >> $BASH_ENV
+              echo "export MNE_3D_OPTION=true" >> $BASH_ENV
               echo "export PATH=~/.local/bin/:$PATH" >> $BASH_ENV
               echo "source ~/python_env/bin/activate" >> $BASH_ENV
               mkdir -p ~/.local/bin
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,7 +65,10 @@ jobs:
               echo "export OPENBLAS_NUM_THREADS=4" >> $BASH_ENV
               echo "export XDG_RUNTIME_DIR=/tmp/runtime-circleci" >> $BASH_ENV
               echo "export MNE_FULL_DATE=true" >> $BASH_ENV
-              echo "export MNE_3D_OPTION=true" >> $BASH_ENV
+              echo "export MNE_3D_BACKEND=pyvistaqt" >> $BASH_ENV
+              echo "export MNE_3D_OPTION_MULTI_SAMPLES=1" >> $BASH_ENV
+              echo "export MNE_BROWSER_BACKEND=qt" >> $BASH_ENV
+              echo "export MNE_BROWSER_PRECOMPUTE=false" >> $BASH_ENV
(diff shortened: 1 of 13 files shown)


.circleci/config.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       docker:
         # Use 18.04 rather than 20.04 because MESA 20.0.8 on 18.04 has working
         # transparency but 21.0.3 on 20.04 does not!
-        - image: cimg/base:stable-18.04
+        - image: cimg/base:2023.03-18.04
```
florin-pop, line 16: I'm not sure that this change is needed, I'll revert it and try again.

.github/workflows/linux_pip.yml
```diff
@@ -27,6 +27,7 @@ jobs:
       OPENBLAS_NUM_THREADS: '1'
       PYTHONUNBUFFERED: '1'
       PYTHON_VERSION: '3.10'
+      QT_DEBUG_PLUGINS: '1'
     strategy:
```
florin-pop, line 31: This will be reverted as well before merge.

requirements.txt
```diff
@@ -33,7 +33,7 @@ xlrd
 imageio>=2.6.1
 imageio-ffmpeg>=0.4.1
 traitlets
-pyvista>=0.32,!=0.35.2,!=0.38.0,!=0.38.1,!=0.38.2,!=0.38.3
+pyvista>=0.32,!=0.35.2,!=0.38.0,!=0.38.1,!=0.38.2,!=0.38.3,!=0.38.4,!=0.38.5
```
florin-pop, line 36: I'm not happy with this change. What do you think about pinning pyvista to `>=0.32,<0.38`?

larsoner: Agreed it's ugly, but IIRC the bug is/should be fixed in their `main` so it should work with the next release 🤷

.circleci/config.yml
larsoner: This comment is very old and IIRC 22.04 mesa might actually be okay. Here I'd do whatever we do in MNE-Python's CircleCI config as things are tested/updated much more often there...

larsoner: (you'll need to make sure you have the MNE_3D_OPTION stuff equivalent between the two... IIRC we discovered the transparency bug was due to full-screen antialiasing being enabled or something)

requirements.txt
rob-luke: Let's stick with ugly then

rob-luke:
Thanks @florin-pop it looks good to me. I also looked through the rendered docs and they were correct (including the 3d renders).

Could you please revert the change you wanted to revert and then ping Eric to merge.

.github/workflows/linux_pip.yml
rob-luke: Please leave in draft PR merge while there are things you intend to revert so we don't accidentally merge things that are intended for debugging.

rob-luke (approved):

.github/workflows/linux_pip.yml
rob-luke: I realised I have access to revert the change ;) Done! THanks Florin

rob-luke:
> Could you please revert the change you wanted to revert and then ping Eric to merge.

I remembered I can revert the change myself ;) So we don't need to bug Eric now and I'll merge when all green

rob-luke:
This is absolutely super! Thanks @florin-pop

Issue #471, opened 2022-07-01
Follow MNE-Python and remove pyqt5 and move to pyside6


rob-luke:
To minimise maintenance and ensure compatability, it is best to follow the practices and packages from MNE-Python. So let's move to pyside6 https://github.com/mne-tools/mne-python/pull/10513

rob-luke:
Before starting on this. Please check if MNE-Python is still using PySide6, or have they moved to PySide32 by now? Use whatever is being used in MNE-Python

larsoner:
PyQt6 for pip, still PyQt5 for conda. PySide6 does not work well for REPL interpreter matplotlib interactive mode.

rob-luke:
Simple 😉
But seriously, thanks for the explanation, greatly appreciated!

PR #510, opened 2023-04-27
MAINT: Migrate from PyQt5 to PySide6

florin-pop:
Fixes https://github.com/mne-tools/mne-nirs/issues/471 by:
- replacing `PyQt5` with `pyside6` in `requirements.txt `which is also used by Circle CI / Azure pipelines
- replacing `PyQt6` with `pyside6` for github workflows
- adding a mitigation for ensuring that some tests are not failing due to a QT deprecation warning

---

### DIFF
--- a/mne_nirs/tests/test_examples.py
+++ b/mne_nirs/tests/test_examples.py
@@ -28,8 +28,8 @@ def examples_path():
 
 def run_script_and_check(test_file_path):
     with open(test_file_path) as fid, warnings.catch_warnings():
-        # Ignore deprecation warning caused by setting app.setAttribute(Qt.AA_UseHighDpiPixmaps)
-        # in mne-python
+        # Ignore deprecation warning caused by
+        # app.setAttribute(Qt.AA_UseHighDpiPixmaps) in mne-python
         warnings.filterwarnings("ignore", category=DeprecationWarning)
         return exec(fid.read(), locals(), locals())
 
--- a/mne_nirs/tests/test_examples.py
+++ b/mne_nirs/tests/test_examples.py
@@ -7,6 +7,7 @@
 import os
 import pytest
 import sys
+import warnings
 
 from mne.utils import check_version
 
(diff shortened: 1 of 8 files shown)


mne_nirs/tests/test_examples.py
```diff
@@ -26,7 +27,10 @@ def examples_path():
 
 
 def run_script_and_check(test_file_path):
-    with open(test_file_path) as fid:
+    with open(test_file_path) as fid, warnings.catch_warnings():
+        # Ignore deprecation warning caused by
```
florin-pop, line 31: I added this workaround in order to attempt to fix these kind of [failures](https://dev.azure.com/mne-tools/mne-nirs/_build/results?buildId=24870&view=logs&jobId=8e630eb1-cb11-5b3c-6e81-b13952459f99&j=8e630eb1-cb11-5b3c-6e81-b13952459f99&t=923c8314-bb6c-5f49-b147-9d7b6ecb81f8) which are caused by a warning that originates from https://github.com/mne-tools/mne-python/blob/cecbf0fb1dc919254bb6095db491b2d7c5f68003/mne/viz/backends/_utils.py#L164
Do you have a better suggestion on how to mitigate this? Maybe we could submit a patch to upstream?

larsoner: Yes a fix in MNE-Python would be appreciated!

rob-luke: What should a fix look like for this? I guess we can't remove the code because MNE-Python still needs to support previous versions of Qt? And it feels dangerous to catch and ignore the errors in MNE-Python. Any suggestions @larsoner ?

larsoner: In MNE we should check the Qt version and set this attribute conditionally. Or do whatever the Qt deprecation says we should do for new enough versions (set a different attribute, no need to set anything, etc. -- I haven't read what they say to do yet)

larsoner: ... and here you could conditionally catch warnings or not depending on the MNE and/or At version if you want

PR #524, opened 2023-10-12
fix: change ci param to errorbar in sns-catplot

Busato:
Two tests were failing because of a deprecation of the `ci` parameter in `sns.catplot` [function](https://seaborn.pydata.org/generated/seaborn.catplot.html).

image

They were changed to `errorbar` instead that do the same thing when using None (both of the were using it)

### DIFF
--- a/doc/changelog.rst
+++ b/doc/changelog.rst
@@ -17,6 +17,7 @@ To install a specific version of the library you would run ``pip install mne-nir
 v0.6.dev0
 ---------
 
+* Changed ci param to errorbar in sns-catplot functions. By `Nicolas Busato`_.
 * Fix bug in SNIRF writer that caused incorrect duration to be written to file. By `Robert Luke`_.
 * Add option to export montage location in SNIRF using the landmarkLabels field. By `Robert Luke`_.
 * Fix continuous integration issues and update test infrastructure. By `Florin Pop`_.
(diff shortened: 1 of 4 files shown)


larsoner:
CI failures appear unrelated, let me fix those in a separate PR then merge `main` into this branch, then should be good to go. Thanks in advance @Busato !

doc/changelog.rst
```diff
@@ -197,3 +197,4 @@ Enhancements
 .. _Samuel Powell: https://github.com/samuelpowell
 .. _Johann Benerradi: https://github.com/HanBnrd
 .. _Florin Pop: https://github.com/florin-pop
+.. _Nicolas Busato: https://github.com/Busato
```
larsoner, line 200: Can you also add a line under the `v0.6.dev0` heading toward the top?

larsoner:
You can now also remove this line that I added https://github.com/mne-tools/mne-nirs/pull/525/files#diff-15c86dcddb07192eeb2d84db2a5dfca48e3837471adbbd9b3e05425b44fce485R49 if you want to make sure it has been fixed everywhere

doc/changelog.rst
Busato: @larsoner do you mean a line saying what the PR changed?

larsoner: Yep!

rob-luke:
Thanks @Busato 🎉

PR #527, opened 2023-12-07
FIX: Fixes for release

larsoner:
Pushed to PyPI and conda-forge, but let's fix the website. I don't think we need to archive old versions like we do for MNE-Python so let's get rid of them.


larsoner:
Changed my mind, it's easy enough to keep the old versions around

PR #572, opened 2024-08-13
put the theme we actually use as a dependency

drammock:
cf: https://github.com/mne-tools/mne-python/issues/12789


larsoner:
Ideally I think the `requirements_doc.txt`, `pyproject.toml`, and `.circleci` (whatever it uses if not one of those) would be deduplicated at some point, but this is already a good enough fix :+1:

PR #586, opened 2024-11-13
initial draft of new TD-fNIRS KF2 example

JohnGriffiths:
Following from [this PR discussion](https://github.com/mne-tools/mne-python/pull/11064), this example demonstrates I/O of TD-FNIRS data, and also serves as an introduction to KF2 data. 

Code and data based on example code from @julien-dubois-k

Remaining to-do:

- [x]  Modify installation target to the above-linked PR 
- [x]  Add data download functionality
- [x]  Modify some of the `hdf` access bits with references to getting the info from the mne `snirf` loader


JohnGriffiths:
PR adding KF2 tutorial example, as [discussed](https://github.com/mne-tools/mne-python/pull/11064#issuecomment-2463258887). 

I have verified that this builds an correct-looking example webpage with `Make html` 

@larsoner  - do need some help here with getting the CI to pass

Some things to note re: the contents of this example:

1. It currently uses `gdown` to pull the example file. I suggest we keep this as-in, and consider updating to something with `mne-misc-data` after this one is merged

2. Likewise, I suggest we edit the `mne` dependency back to the `mne-tools/mne-python` master in a later PR, once the PR in the above-linked thread has been resolved. 

3.  Some of the contents regarding pulling stuff from the hdf5 archive could be modified in future versions to show pulling this info from the hdf archive directly or from the mne reader. However I don't believe this functionality is in the MNE `.snirf` reader yet, so we would need to complete a new PR with that code first.

tools/circleci_dependencies.sh
```diff
@@ -11,6 +11,7 @@ echo "Installing dependencies"
 python -m pip install --upgrade --progress-bar off \
     -r requirements.txt -r requirements_testing.txt -r requirements_doc.txt \
     "mne-qt-browser[opengl] @ git+https://github.com/mne-tools/mne-qt-browser.git@main" \
+    "mne-python @ git+https://github.com/larsoner/mne-python.git@feature/TD-nirs_snirf" \
```
larsoner, line 14: The forward slash is causing problems, let's try this
```suggestion
    "https://github.com/larsoner/mne-python/archive/refs/heads/feature/TD-nirs_snirf.zip" \
```

larsoner:
FYI I am a bit busy for the next two weeks then should be able to come back to this!

larsoner:
```
generating gallery for auto_examples/general... [ 27%] plot_11b_kf2_fingertapping.py
make: *** [Makefile:60: html] Killed
```
Probably due to memory usage

Screenshot 2024-11-18 at 11 39 57 AM

Is there some easy / reasonable way to reduce the memory usage? It's good if examples don't use a ton of memory (e.g., try to keep to < 2 GB).

In a worst case we could use a larger docker image but it would be nice to avoid it if possible

JohnGriffiths:
> Is there some easy / reasonable way to reduce the memory usage? It's good if examples don't use a ton of memory (e.g., try to keep to < 2 GB)

If it's an option to keep as-is with larger docker image can we go with that for now?

It would be helpful to focus on the code needed to get this up and running first, and circle back to optimizations next. 

I think RAM usage limits are also a more general question that would inform a lot of what we decide to add examples for on this topic (e.g. connectivity, source analysis, etc.).

larsoner:
> It would be helpful to focus on the code needed to get this up and running first, and circle back to optimizations next.

Okay I pushed a commit that should do it.

FWIW bumping the resource class actually costs the mne-tools org compute credits because it's not part of the free tier so we should try to avoid keeping it that way as well long-term if possible unless there is a really compelling use case that we can't get memory usage down for. Even in MNE-Python so far we have managed to avoid this, only in MNE-BIDS-Pipeline have we needed to use it. So I'm cautiously optimistic we might be able to keep memory usage low here, too.

examples/general/plot_11b_kf2_fingertapping.py
```diff
@@ -0,0 +1,437 @@
...
+    if idx_time == 0:
+        ax[0][0].set_ylabel("LEFT")
+        ax[1][0].set_ylabel("RIGHT")
+        ax[2][0].set_ylabel("LEFT - RIGHT")
+fig.suptitle(chromophore)
+
+
+# %%
+#
+# Despite the absence of thresholding, we can discern:
```
larsoner, line 277:
```suggestion
# Despite the absence of thresholding, we can discern:

```

JohnGriffiths:
> > It would be helpful to focus on the code needed to get this up and running first, and circle back to optimizations next.
> 
> Okay I pushed a commit that should do it.
> 
> FWIW bumping the resource class actually costs the mne-tools org compute credits because it's not part of the free tier so we should try to avoid keeping it that way as well long-term if possible unless there is a really compelling use case that we can't get memory usage down for. Even in MNE-Python so far we have managed to avoid this, only in MNE-BIDS-Pipeline have we needed to use it. So I'm cautiously optimistic we might be able to keep memory usage low here, too.

Good, thanks. 

Agree to avoiding this being a long-term solution.

JohnGriffiths:
Snooped around and found the [updated rendered example](https://output.circle-artifacts.com/output/job/5232ce6b-f773-4c77-ab66-4d695a1c3382/artifacts/0/html/auto_examples/general/plot_11b_kf2_fingertapping.html) from the above small tweak - looks to have worked as intended.

JohnGriffiths:
Hi @larsoner @julien-dubois-k 

Sorry for slow turnaround but a few interim updates for this thread:

I just collected some fresh finger tapping recording data with KF2, so that we could look at the moments as well as Hb, as @larsoner is requesting. These are now saved and available with recordings, but weren't for the earlier example dataset. I thought it would also be good to get a more up-to-date example dataset anyway. 

I will do an update on the example code in this PR using that data in the next few days. Judging from the kernel portal auto-pipelines figures I think this should be a good example dataset

t-stats for right-handed finger tapping > fixation cross (thresh at sig p<0.001)
right_vs_fixation_HbO_p001

t-stats for left-handed finger tapping > fixation cross (thresh at sig p<0.001)
left_vs_fixation_HbO_p001

I will have a go at updating the example for both the pre-computed HbO and a corresponding analysis starting from the moments, which I believe is what @larsoner is wanting to do. 

I will also do a parallel PR to MNE-Python with a data fetcher for this dataset, as opposed to the gdrive download used so far. I created an [OSF repo](https://osf.io/6xhwt/) for this, which I will populate with the above data shortly. @larsoner lmk if you would prefer to do that differently.

larsoner:
Sounds like a good plan @JohnGriffiths !

larsoner:
Okay I moved the dataset downloader here and got CIs green. @JohnGriffiths can you look into reducing memory consumption and runtime if possible?

JohnGriffiths:
Yes I can look into memory stuff. 

LMK

- What is the tool you use to generate this plot?
- What are the run time and memory usage levels to aim for?
- Are you waiting on these tweaks to merge or can we do that as a follow-up PR?

drammock:
> What is the tool you use to generate this plot?

`memory_profiler` is the tool. Run on command line like:

```
mprof run python my_script.py
mprof plot
```

> What are the run time and memory usage levels to aim for?

off the cuff: less than a minute and less than 2GB (1GB would be better) is what we shoot for in the MNE-Python repo. @larsoner may have a clearer picture of what is OK for MNE-NIRS.

> Are you waiting on these tweaks to merge or can we do that as a follow-up PR?

I'd say we're waiting; if we merge now, then every PR will use the large resource class (costing us more money) until the usage is decreased and the resource class reverted to default.

JohnGriffiths:
Looks fine to me. Nice trick!

larsoner:
... and version numbers/releases are pretty easy nowadays, here you go:

https://pypi.org/project/mne-nirs/0.7.3/

Should land in `conda` soon as well!

JohnGriffiths:
> Okay in it goes, then! Would be good to work on the moments next when you get a chance :smile: 

Ok yes moments next 

And also DOT analysis for this data


> 
> > Are there any mne-nirs tutorial slide decks etc. you could point me to / send to me that I can borrow from?
> 
> No not that I know of... back when I gave MNE-Python tutorials I'd generally start from the `ipynb` Notebook files from the tutorials themselves and build out from there depending on what the group(s) in particular wanted to learn


Yup ok this is basically my plan

Issue #413, opened 2021-11-16
Multicollinearity


rob-luke:
I recently saw an insightful presentation from @helenacockx which described how multicollinearity/collinearity ([wikipedia](https://en.wikipedia.org/wiki/Multicollinearity)) can affect fNIRS GLM analysis. MNE-NIRS provides support for both averaging and GLM analysis. The issue of collinearity in fNIRS GLM analysis has been discussed in the literature, "a challenge of these regressor models is collinearity introduced between the task and nuisance regressors, which can happen if the systemic physiological response is correlated with the performance of the task. Collinearity in the regression analysis can destabilize it due to poor mathematical conditioning of the model and can produce unpredictable results." (Santosa et al 2020). Specifically, this is of concern as short channels (designed to measure systemic activity, not neural activity) included in the design matrix to remove systemic contributions to the signal can be highly correlated with the task regressor.

Relevant reading:
* https://dartbrains.org/content/GLM_Single_Subject_Model.html#multicollinearity
* Santosa 2020 used PCA to remove colinearity between the nuisance regressors, but this did not solve the issue of colinearity between short channels and the task regressor. "This was only used to remove collinearity from using multiple short-seperation channels in the model, but did not reduce any collinearity between the task regressors and the short-seperation nuisance regressors."
* Need to find more articles on the topic

#### Tasks

- [ ] Add metrics to quantify colinearity in the design matrix
- [ ] Implement methods to mitigate the problem of colinearity 

#### References

Hendrik Santosa, Xuetong Zhai, Frank Fishburn, Patrick J. Sparto, Theodore J. Huppert, "Quantitative comparison of correction techniques for removing systemic physiological signal in functional near-infrared spectroscopy studies," Neurophoton. 7(3) 035009 (23 September 2020)

helenacockx:
Thanks for picking this up Rob! The article of Santosa is indeed the only fNIRS-related paper that I found on this topic. In this paper, they propose to solve the issue of collinearity by performing a regularized mixed-effects model estimation. However, they concluded that the mixed-effect model only showed a slight improvement in performance (type-I errors) compared to the AR-IRLS model, this came at a cost of a 10-fold computation time.

I found that this article also gave insight into the concept of multicollinearity and why not solve it with orthogonalization: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4412813/

[...]

PR #676, opened 2026-03-25
adds vif metric from stats models

kikiluvbrains:
#### Reference issue
Multi-collinearity #[413](https://github.com/mne-tools/mne-nirs/issues/413).

Add metrics to quantify collinearity in the design matrix. One simple way to deal with high multi-collinearity (typically VIF > 4–5) is to combine very similar regressors or drop the ones that are problematic.

There are other ways to handle high VIF, such as using principal component analysis (PCA), but I haven’t explored those yet, I am open to suggestions.


larsoner (commented):
Looks like a good start!

mne_nirs/experimental_design/_experimental_design.py
```diff
@@ -4,6 +4,8 @@
 
 import mne
 import numpy as np
+from mne.utils import logger
+from statsmodels.stats.outliers_influence import variance_inflation_factor
```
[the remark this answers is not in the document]
kikiluvbrains: Hey, thank you so much for your feedback!

I think we should opt out of statsmodels, I  wrote an equivalent test for variance_inflation_factor, and also verified with some test data which gave me equivalent results to statsmodels functionality. 

let me know what you think.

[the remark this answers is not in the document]
larsoner: Yeah if it's simple enough feel free to do it that way

You could even write a little helper test that uses `statsmodels` (only in the test!) to verify equivalence of the helper function if you want (and then have CIs install it just for testing)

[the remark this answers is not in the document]
kikiluvbrains: sounds good, I will add in a little helper test

larsoner:
Let me know when I should look!

kikiluvbrains:
@larsoner I am ready for you to take a look, currently am trying to pass all checks

kikiluvbrains:
@larsoner 

I do have an optional flag for "vif_export" as a way to test against stasmodel, but it causes this error in some of the html examples as shown in circleCI, let me know what you think, and best way I can patch this?

```
        ../examples/general/plot_11b_kf2_fingertapping.py failed leaving traceback:
    
        Traceback (most recent call last):
          File "/home/circleci/project/examples/general/plot_11b_kf2_fingertapping.py", line 416, in <module>
            design_matrix = make_first_level_design_matrix(
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
          File "/home/circleci/project/mne_nirs/experimental_design/_experimental_design.py", line 106, in make_first_level_design_matrix
            dm = make_first_level_design_matrix(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        TypeError: make_first_level_design_matrix() got an unexpected keyword argument 'vif_export'
```

mne_nirs/experimental_design/tests/test_experimental_design.py
```diff
@@ -7,6 +7,9 @@
 import mne
 import numpy as np
 
+# for the comparison of vif we need these two libraries
+from statsmodels.stats.outliers_influence import variance_inflation_factor
```
larsoner: This should be imported inside the test in case people don't have it installed, and in the test you can hav.
```
pytest.importorskip("statsmodels")
```

mne_nirs/experimental_design/tests/test_experimental_design.py
```diff
@@ -123,3 +181,30 @@ def test_high_pass_helpers():
...
+
+    design_matrix_statsm, vif_statsm = make_first_level_design_matrix_w_statsmodels_vif(
+        raw_intensity, drift_order=1, drift_model="polynomial"
+    )
+
+    # expect near identical results but not exact since ourrs is using glm from nii.learn
+    # wheras statsmodel has their own implmentation before extracting the vif values
+    # note vif will come with a level of uncertainity +/- 0.05 of what is reported
+    for key in vif:
+<<<<<<< HEAD
```
larsoner: Looks like some problem with rebasing

mne_nirs/experimental_design/_experimental_design.py
```diff
@@ -73,16 +74,22 @@ def make_first_level_design_matrix(
     oversampling : int, optional
         Oversampling factor used in temporal convolutions. Default=50.
 
+    vif export : bool, optional
+        deafult set to false, if set to True will export vif values;
```
larsoner: Hmm... this probably shouldn't be part of the public API, so let's remove it

If we really had to have a way to get the values for the test we would want to create a private `_make_first_level_design_matrix` with this option available and the public `make_first_level_design_matrix` would always call it with `return_vif=False`, and the test could use the private one with `return_vif=True`

mne_nirs/experimental_design/_experimental_design.py
```diff
@@ -108,9 +115,47 @@ def make_first_level_design_matrix(
...
+        vif = 1.0 / (1.0 - r_squared_i)
+        vif_all.append(vif)
+
+    # alert user if high vif detected
+    for name, vif_idx in zip(predictor_names, vif_all):
+        msg = f"{name} with VIF of {vif_idx:.3f}"
+        if vif_idx > 4:
+            logger.warning("High collinearity " + msg)
+        else:
+            logger.info(msg)
```
larsoner: You are already logging the `VIF`s so in the test you can do something like the following to recover the values:
```
from mne.utils import use_log_level, catch_logging

...
def some_test():
    ...
    with use_log_level("info"), catch_logging() as log:
        ... make_first_level_design_matrix(...)
    log = log.getvalue()
    vifs = np.array([line.split()[-1] for line in log.splitlines() if " VIF " in line], float)
```

mne_nirs/experimental_design/tests/test_experimental_design.py
kikiluvbrains: will do

kikiluvbrains:
Sorry for the delay! Everything has been updated per your suggestions. Let me know what you think when you get a chance.

kikiluvbrains:
Hi! fyi I just force-pushed a cleaned-up history to this branch so all the changes are squashed into a single commit (had to go back and forith a bit with CircleCI)

larsoner (approved):
I implemented a simplification and unified into a single function with` return_vif`. Marking for merge-when-green. Thanks in advance @kikiluvbrains !

kikiluvbrains:
Hi, thanks for this! Having mne-nirs's own VIF utility tools is very helpful in case any functions from Ni-learn ever becomes deprecated in the near future.

PR #680, opened 2026-04-13
ENH/FIX: Multi-wavelength support for scalp_coupling_index_windowed, peak_power, and plot_3d_montage

### LINKED ISSUE #631: Support for Multi-Wavelength (>2) NIRS/SNIRF Data Processing
#### Describe the new feature or enhancement
Goal: extend MNE-Python and MNE-NIRS to support an arbitrary number of wavelengths (≥2) instead of being limited (hard-coded in several places) to two.

The majority of changes to achieve this is in MNE-Python, see the issue https://github.com/mne-tools/mne-python/issues/13401 that I opened there. The idea was also mentioned in https://github.com/mne-tools/mne-python/issues/9816.

Before submitting a PR with the changes, I'd like to discuss implementation details and to explore devs' opinion on how feasible such changes would be.

#### Describe your proposed implementation
I think the changes only affect two files; I haven't looked at whether tests need to be updated or not.

Changes:
- [`write_raw_snirf()`](https://github.com/mne-tools/mne-nirs/blob/main/mne_nirs/io/snirf/_snirf.py#L19) refers to the read function in MNE-python in the docs. Once the read function can handle >=2 wavelengths, the doc will need to reflect that.
- [/mne_nirs/preprocessing/_scalp_coupling_segmented.py: `_validate_nirs_info`](https://github.com/mne-tools/mne-nirs/blob/main/mne_nirs/preprocessing/_scalp_coupling_segmented.py#L13) is a windowed version of a similar function in mne-python. I suggested there, that for >=2 wavelength, the the minimum of all correlation pairs should be used. My proposed implementation here also includes some changes to make the code more similar to mne-python.

A preliminary, untested implementation of the proposed code changes can be seen at [my fork](https://github.com/mne-tools/mne-nirs/compare/main...zEdS15B3GCwq:mne-nirs:more_wavelengths).

#### Describe possible alternatives
Maybe other ways to calculate SCI for >=2 wavelengths. It's probably best to keep SCI calculation in sync with mne-python's non-windowed implementation, so I suggest to discuss this in https://github.com/mne-tools/mne-python/issues/13401 .

zEdS15B3GCwq:
## Summary

Adds multi-wavelength (≥2) support to `scalp_coupling_index_windowed`,
`peak_power`, and `plot_3d_montage`, following the approach established in
mne-tools/mne-python#13408. Also fixes two bugs in the preprocessing functions
and one in the visualisation function that assumed a specific channel ordering,
which does not hold in general.

Closes #631.

### `mne_nirs/preprocessing/_scalp_coupling_segmented.py` (ENH + FIX)

- The function previously hard-coded 2-wavelength pair selection using `[::2]` /
  `[1::2]` index strides. It now uses `_channel_frequencies` (imported from
  `mne.preprocessing.nirs`) to determine the number of distinct wavelengths, and
  iterates over channel groups of that size.
- For `n_wavelengths > 2`, all pairwise correlations within the group are computed
  and the **minimum** is used as the SCI score — matching the mne-python approach.
- **Bug fix:** Bad-channel annotations (`BAD_SCI`) were previously assigned using
  sequential indices (0, 1, 2, …) rather than the actual channel indices returned
  by `_validate_nirs_info`. Because `picks` can be in a different order than
  `raw.ch_names`, this caused annotations to be attached to the wrong channels
  when channel order was non-alphabetical (as is common in practice with NIRx and
  Shimadzu data).

### `mne_nirs/preprocessing/_peak_power.py` (ENH + FIX)

- Same multi-wavelength generalisation as SCI: `_channel_frequencies`-based group
  size, pairwise cross-correlation + periodogram, **minimum** peak power per group.
- **Bug fix:** Same incorrect annotation assignment as SCI — fixed identically.

### `mne_nirs/visualisation/_plot_3d_montage.py` (FIX)

- The function selected one representative channel per source-detector pair using
  `picks[::2]`, hard-coding a stride of 2. This broke with ≥3 wavelengths and with
  any dataset where channels were not in the expected interleaved order.
- Fixed by using `np.unique(prefixes, return_index=True)` to select the first
  occurrence of each unique source-detector prefix, independent of stride and order.

### `mne_nirs/preprocessing/tests/test_quality.py`

- Added `fnirs_labnirs_3wl_data` fixture backed by the Shimadzu Labnirs 3-wavelength
  SNIRF file now available in `mne-testing-data`.
- Added `test_sci_windowed_known_values_multi_wavelength` and
  `test_peak_power_known_values_multi_wavelength` — synthetic known-value tests for
  3-wavelength data covering correct score computation and annotation placement.
- Added `test_sci_windowed_annotations_target_correct_channels` and
  `test_pp_windowed_annotations_target_correct_channels` — regression tests that
  construct a minimal 4-channel dataset with non-alphabetical channel order and
  verify that bad annotations target the correct channels.

## Notes

- The algorithmic differences from Pollonini's original (signal type, window
  length, threshold defaults, cosine similarity vs. Pearson correlation) noted in
  #631 are **out of scope** for this PR.
- Same goes for tests for expected behaviour of SCI and PP measures
  (e.g. scale invariability).
- The 3-wavelength test data requires `mne-testing-data`; tests are guarded with
  `@mne.datasets.testing.requires_testing_data`.

### DIFF
--- a/mne_nirs/preprocessing/_peak_power.py
+++ b/mne_nirs/preprocessing/_peak_power.py
@@ -56,7 +56,7 @@ def peak_power(
 
     Notes
     -----
-    This implementation of peak power differs from Pollonini's original[1]_[2]_,
+    This implementation of peak power differs from Pollonini's original [1]_ [2]_,
     in that the original calculates peak power on raw data, whereas multiple
     types are allowed here; and while both implementations calculate a kind of
     cosine similarity, the mathematical details are different. Users are
--- a/mne_nirs/preprocessing/_peak_power.py
+++ b/mne_nirs/preprocessing/_peak_power.py
@@ -54,13 +54,13 @@ def peak_power(
         List of the start and end times of each window used to compute the
         peak spectral power.
 
-    Warning
-    -------
-    This implementation of peak power differs from Pollonini's original in that
-    the original calculates peak power on raw data, whereas multiple types are
-    allowed here; and while both implementations calculate a kind of cosine
-    similarity, the mathematical details are different. Users are advised to
-    check the results and adjust parameters as needed.
+    Notes
+    -----
+    This implementation of peak power differs from Pollonini's original[1]_[2]_,
+    in that the original calculates peak power on raw data, whereas multiple
+    types are allowed here; and while both implementations calculate a kind of
+    cosine similarity, the mathematical details are different. Users are
+    advised to check the results and adjust parameters as needed.
 
     References
     ----------
--- a/mne_nirs/preprocessing/_peak_power.py
+++ b/mne_nirs/preprocessing/_peak_power.py
@@ -5,7 +5,7 @@
 import numpy as np
 from mne.filter import filter_data
 from mne.io import BaseRaw
-from mne.preprocessing.nirs import _validate_nirs_info
+from mne.preprocessing.nirs import _channel_frequencies, _validate_nirs_info
 from mne.utils import _validate_type, verbose
 from scipy.signal import periodogram
 
@@ -54,6 +54,14 @@ def peak_power(
         List of the start and end times of each window used to compute the
         peak spectral power.
 
+    Warning
+    -------
+    This implementation of peak power differs from Pollonini's original in that
+    the original calculates peak power on raw data, whereas multiple types are
+    allowed here; and while both implementations calculate a kind of cosine
+    similarity, the mathematical details are different. Users are advised to
+    check the results and adjust parameters as needed.
+
     References
     ----------
     .. [1] Pollonini L et al., “PHOEBE: a method for real time mapping of
@@ -63,11 +71,24 @@ def peak_power(
            quality assessment of fNIRS scans." Optics and the Brain.
            Optical Society of America, 2020.
     """
+    # Copy raw to avoid modifying original and load data into memory
     raw = raw.copy().load_data()
+
+    # Validate that the input contains raw fNIRS data
+    # Note that peak_power currently does not require a specific data type (e.g. OD)
     _validate_type(raw, BaseRaw, "raw")
 
+    # `picks` returns a list of channels ordered alphanumerically and grouped by
+    # channel groups (i.e. same source-detector pair). The order of channels in
+    # `picks` can be differ from the order of channels in `raw`.
     picks = _validate_nirs_info(raw.info)
 
+    # Number of wavelengths extracted from channel names
+    n_wavelengths = len(np.unique(_channel_frequencies(raw.info)))
+
+    # Bandpass filter data to extract heartbeat-related frequencies
+    # Note: filtering is applied only to the selected channels (picks),
+    # with channel order preserved, regardless of how the picks are ordered.
     filtered_data = filter_data(
         raw._data,
         raw.info["sfreq"],
@@ -79,42 +100,49 @@ def peak_power(
         h_trans_bandwidth=h_trans_bandwidth,
     )
 
-    window_samples = int(np.ceil(time_window * raw.info["sfreq"]))
-    n_windows = int(np.floor(len(raw) / window_samples))
+    samples_per_window = int(np.ceil(time_window * raw.info["sfreq"]))
+    n_windows = int(np.floor(len(raw) / samples_per_window))
 
     scores = np.zeros((len(picks), n_windows))
     times = []
 
     for window in range(n_windows):
-        start_sample = int(window * window_samples)
-        end_sample = start_sample + window_samples
-        end_sample = np.min([end_sample, len(raw) - 1])
+        start_sample = int(window * samples_per_window)
+        end_sample = min(start_sample + samples_per_window, len(raw) - 1)
 
         t_start = raw.times[start_sample]
         t_stop = raw.times[end_sample]
         times.append((t_start, t_stop))
 
-        for ii in range(0, len(picks), 2):
-            c1 = filtered_data[picks[ii]][start_sample:end_sample]
-            c2 = filtered_data[picks[ii + 1]][start_sample:end_sample]
+        # pair indices for all channels pairs
+        pair_indices = np.triu_indices(n_wavelengths, k=1)
 
-            # protect against zero
-            c1 = c1 / (np.std(c1) or 1)
-            c2 = c2 / (np.std(c2) or 1)
+        for gg in range(0, len(picks), n_wavelengths):
+            ch_group = picks[gg : gg + n_wavelengths]
+            group_data = filtered_data[ch_group, start_sample:end_sample]
 
-            c = np.correlate(c1, c2, "full")
-            c = c / (window_samples)
-            [f, pxx] = periodogram(c, fs=raw.info["sfreq"], window="hamming")
+            # Calculate pairwise peak power within group
+            group_data = np.array([ch / (np.std(ch) or 1) for ch in group_data])
+            peak_powers = []
+            for ii, jj in zip(*pair_indices):
+                c = np.correlate(group_data[ii], group_data[jj], "full")
+                c = c / samples_per_window
+                [_, pxx] = periodogram(c, fs=raw.info["sfreq"], window="hamming")
+                peak_powers.append(max(pxx))
 
-            scores[ii, window] = max(pxx)
-            scores[ii + 1, window] = max(pxx)
+            # Use the minimum value in the group as peak power
+            pp = min(peak_powers) if peak_powers else 0.0
 
-            if (threshold is not None) & (max(pxx) < threshold):
+            # Assign the same peak power value to all channels in the group
+            scores[ch_group, window] = pp
+
+            # Add BAD_PeakPower annotation to channels if below threshold
+            if (threshold is not None) & (pp < threshold):
                 raw.annotations.append(
                     t_start,
                     time_window,
                     "BAD_PeakPower",
-                    ch_names=[raw.ch_names[ii : ii + 2]],
+                    ch_names=[[raw.ch_names[ii] for ii in ch_group]],
                 )
-    scores = scores[np.argsort(picks)]
+
     return raw, scores, times
--- a/mne_nirs/preprocessing/_peak_power.py
+++ b/mne_nirs/preprocessing/_peak_power.py
@@ -78,9 +78,11 @@ def peak_power(
     # Note that peak_power currently does not require a specific data type (e.g. OD)
     _validate_type(raw, BaseRaw, "raw")
 
-    # `picks` returns a list of channels ordered alphanumerically and grouped by
-    # channel groups (i.e. same source-detector pair). The order of channels in
-    # `picks` can be differ from the order of channels in `raw`.
+    # `picks` returns a list of channels ordered alphanumerically, which may differ
+    # from the order of channels in `raw`. By virtue of being sorted, channels follow
+    # an ordered sequence of S-D pairs and wavelengths, e.g., S1_D1 760, S1_D1 850,
+    # S1_D2 760, S1_D2 850, S2_D1 760, S2_D1 850, etc. The algorithm below relies on
+    # this ordering.
     picks = _validate_nirs_info(raw.info)
 
     # Number of wavelengths extracted from channel names
--- a/mne_nirs/preprocessing/_scalp_coupling_segmented.py
+++ b/mne_nirs/preprocessing/_scalp_coupling_segmented.py
@@ -55,7 +55,7 @@ def scalp_coupling_index_windowed(
     Notes
     -----
     This implementation of scalp coupling index differs from Pollonini's
-    original[1]_[2]_, in that the original calculates scalp coupling index on raw
+    original [1]_ [2]_, in that the original calculates scalp coupling index on raw
     data, whereas optical density data is expected here; and while both
     implementations calculate a kind of cosine similarity, the mathematical
     details are different. Users are advised to check the results and adjust
--- a/mne_nirs/preprocessing/_scalp_coupling_segmented.py
+++ b/mne_nirs/preprocessing/_scalp_coupling_segmented.py
@@ -77,9 +77,11 @@ def scalp_coupling_index_windowed(
     _validate_type(raw, BaseRaw, "raw")
 
     # Pick optical density channels
-    # `picks` returns a list of channels ordered alphanumerically and grouped by
-    # channel groups (i.e. same source-detector pair). The order of channels in
-    # `picks` can be differ from the order of channels in `raw`.
+    # `picks` returns a list of channels ordered alphanumerically, which may differ
+    # from the order of channels in `raw`. By virtue of being sorted, channels follow
+    # an ordered sequence of S-D pairs and wavelengths, e.g., S1_D1 760, S1_D1 850,
+    # S1_D2 760, S1_D2 850, S2_D1 760, S2_D1 850, etc. The algorithm below relies on
+    # this ordering.
     picks = _validate_nirs_info(raw.info, fnirs="od", which="Scalp coupling index")
 
     # Number of wavelengths extracted from channel names
--- a/mne_nirs/preprocessing/_scalp_coupling_segmented.py
+++ b/mne_nirs/preprocessing/_scalp_coupling_segmented.py
@@ -52,13 +52,14 @@ def scalp_coupling_index_windowed(
     times : list
         List of the start and end times of each window used to compute SCI.
 
-    Warning
-    -------
-    This implementation of scalp coupling index differs from Pollonini's original in
-    that the original calculates scalp coupling index on raw data, whereas optical
-    density data is expected here; and while both implementations calculate a kind
-    of cosine similarity, the mathematical details are different. Users are advised
-    to check the results and adjust parameters as needed.
+    Notes
+    -----
+    This implementation of scalp coupling index differs from Pollonini's
+    original[1]_[2]_, in that the original calculates scalp coupling index on raw
+    data, whereas optical density data is expected here; and while both
+    implementations calculate a kind of cosine similarity, the mathematical
+    details are different. Users are advised to check the results and adjust
+    parameters as needed.
 
     References
     ----------
--- a/mne_nirs/preprocessing/_scalp_coupling_segmented.py
+++ b/mne_nirs/preprocessing/_scalp_coupling_segmented.py
@@ -5,7 +5,7 @@
 import numpy as np
 from mne.filter import filter_data
 from mne.io import BaseRaw
-from mne.preprocessing.nirs import _validate_nirs_info
+from mne.preprocessing.nirs import _channel_frequencies, _validate_nirs_info
 from mne.utils import _validate_type, verbose
 
 
@@ -21,7 +21,7 @@ def scalp_coupling_index_windowed(
     verbose=False,
 ):
     """
-    Compute scalp coupling index for each channel and time window.
+    Compute scalp coupling index (SCI) for each channel and time window.
 
     As described in [1]_ and [2]_.
     This method provides a metric of data quality along the duration of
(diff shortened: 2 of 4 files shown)


zEdS15B3GCwq:
@larsoner, better late than never, sorry for the huge delay. Picking this up from our discussion in #631.
Happy to hear any thoughts on the approach.

mne_nirs/preprocessing/tests/test_quality.py
```diff
@@ -2,28 +2,551 @@
 #
 # License: BSD (3-clause)
 
-import os
+from pathlib import Path
```
kikiluvbrains, line 5: is pathlib natively supported by mne?

kikiluvbrains:
@zEdS15B3GCwq could you register your github under circleci and then rerun the build? I want to see weather you would need to make any changes to "make html" on the examples there that use _peak_power.py

mne_nirs/preprocessing/_peak_power.py
```diff
@@ -54,6 +54,14 @@ def peak_power(
         List of the start and end times of each window used to compute the
         peak spectral power.
 
+    Warning
+    -------
+    This implementation of peak power differs from Pollonini's original in that
+    the original calculates peak power on raw data, whereas multiple types are
```
kikiluvbrains: does your implementation allow the option for a user to opt for Pollonini's original implementation

mne_nirs/preprocessing/_peak_power.py
```diff
@@ -79,42 +100,49 @@ def peak_power(
...
-            c1 = c1 / (np.std(c1) or 1)
-            c2 = c2 / (np.std(c2) or 1)
+            # Calculate pairwise peak power within group
+            group_data = np.array([ch / (np.std(ch) or 1) for ch in group_data])
+            peak_powers = []
+            for ii, jj in zip(*pair_indices):
+                c = np.correlate(group_data[ii], group_data[jj], "full")
+                c = c / samples_per_window
+                [_, pxx] = periodogram(c, fs=raw.info["sfreq"], window="hamming")
+                peak_powers.append(max(pxx))
```
kikiluvbrains, line 133: so you taking peak power across all channel pairs?

mne_nirs/preprocessing/_peak_power.py
```diff
@@ -79,42 +100,49 @@ def peak_power(
...
-            c = np.correlate(c1, c2, "full")
-            c = c / (window_samples)
-            [f, pxx] = periodogram(c, fs=raw.info["sfreq"], window="hamming")
+            # Use the minimum value in the group as peak power
+            pp = min(peak_powers) if peak_powers else 0.0
 
-            scores[ii, window] = max(pxx)
-            scores[ii + 1, window] = max(pxx)
+            # Assign the same peak power value to all channels in the group
+            scores[ch_group, window] = pp
```
kikiluvbrains, line 139: so this is the minimum peak power across all pairs for a certain channel pick and chosen window.
you allow for many windows and many channel picks to pair against I assume

mne_nirs/preprocessing/_peak_power.py
```diff
@@ -79,42 +100,49 @@ def peak_power(
...
                 raw.annotations.append(
                     t_start,
                     time_window,
                     "BAD_PeakPower",
-                    ch_names=[raw.ch_names[ii : ii + 2]],
+                    ch_names=[[raw.ch_names[ii] for ii in ch_group]],
                 )
-    scores = scores[np.argsort(picks)]
+
     return raw, scores, times
```
kikiluvbrains, line 150: this will be giving you times for all window pairs, as well as with scores, you should check with circleCI  if this new output will break any existing examples using _peak_powers.py

kikiluvbrains: since now scores will have values for all channel wavelengths, before the no. of wavelengths was harcd-oded to 2

mne_nirs/preprocessing/_peak_power.py
```diff
@@ -63,11 +71,24 @@ def peak_power(
...
 
+    # `picks` returns a list of channels ordered alphanumerically and grouped by
+    # channel groups (i.e. same source-detector pair). The order of channels in
+    # `picks` can be differ from the order of channels in `raw`.
     picks = _validate_nirs_info(raw.info)
 
+    # Number of wavelengths extracted from channel names
+    n_wavelengths = len(np.unique(_channel_frequencies(raw.info)))
+
+    # Bandpass filter data to extract heartbeat-related frequencies
```
kikiluvbrains, line 91: do you do this type of filtering prior to this function call?

mne_nirs/preprocessing/tests/test_quality.py
zEdS15B3GCwq: This is probably a question not addressed to me, but I'd like to hear if there are limitations on including standard libraries. MNE is python >=3.10, which supports pathlib, so I assumed it was OK to use it.

mne_nirs/preprocessing/_peak_power.py
zEdS15B3GCwq: In short, no. In the conversation leading up to this PR (#631), it came up that such a warning could be included. Changing actual behaviour is out of scope for this PR. In a sense, the warning itself is out of scope, so one could argue not to include it at this point.

mne_nirs/preprocessing/_peak_power.py
zEdS15B3GCwq: I'm not sure what you meant by this question. The obvious answer would be no, as the input dataset is not supposed to be filtered and the `filter_data` call is just below, but you probably know that already.

mne_nirs/preprocessing/_peak_power.py
zEdS15B3GCwq: Yes, each pair within the same group.

Both here and in SCI, all pairs of data within one S-D group (i.e. different wavelengths that belong to the same S-D optode pair) are "compared", then the minimum value is used. This is based on the assumption that all channels within a group should be "similar".

MNE-Python's SCI function was also updated to use the same method in the PR https://github.com/mne-tools/mne-python/pull/13408.

mne_nirs/preprocessing/_peak_power.py
zEdS15B3GCwq: The current implementation already iterates over all windows and all picks of the dataset, the only addition is that picks are now grouped according to S-D pairs and within-group minima are taken, and that results are stored in `scores` at the correct indices (`ch_group`), not fixed ones assuming ABAB... channel order.

mne_nirs/preprocessing/_peak_power.py
zEdS15B3GCwq: I'll try the CircleCI thing.

zEdS15B3GCwq:
I've authorised CircleCI, not sure how to trigger a rerun of the tests, so I'll just update the PR with a rebase expecting the tests to run again.

zEdS15B3GCwq:
Fixed some docstring issues that prevented circleCI from succeeding: the `Warning` section was renamed to the more standard `Notes`. `Warnings` (plural) would've worked too, but I felt that _warning_ plural or not, sounded a tad bit too serious. Let me know if you preferred it that way.

By the way, I'd appreciate feedback about the changes to how channel ordering is handled, whether the previous behaviour is actually a bug or I misunderstood how the functions worked.

mne_nirs/preprocessing/tests/test_quality.py
larsoner: Yes `pathlib` is preferred nowadays over `os.path`

mne_nirs/preprocessing/tests/test_quality.py
```diff
@@ -2,28 +2,551 @@
...
+    fnirs_data_folder = Path(mne.datasets.fnirs_motor.data_path())
+    fnirs_raw_dir = fnirs_data_folder / "Participant-1"
+    raw = mne.io.read_raw_nirx(str(fnirs_raw_dir), verbose=True).load_data()
+    return mne.preprocessing.nirs.optical_density(raw)
 
-    raw, scores, times = peak_power(raw)
-    assert len(scores) == len(raw.ch_names)
 
+@pytest.fixture(name="fnirs_labnirs_3wl_data")
+def fixture_fnirs_labnirs_3wl_data() -> mne.io.BaseRaw:
```
larsoner, line 25: Everywhere else in MNE-land we just do
```suggestion
def fnirs_labnirs_3wl_data() -> mne.io.BaseRaw:
```
and know anything inside a `test_whatever(...)` signature is a fixture

mne_nirs/preprocessing/tests/test_quality.py
```diff
@@ -2,28 +2,551 @@
...
+    """Read and return 3-wavelength testing data."""
+    pytest.importorskip("h5py")
```
larsoner: pyproject.toml requires `h5io`, which in turn requires `h5py` so no need for a skipper
```suggestion
```

mne_nirs/preprocessing/tests/test_quality.py
```diff
@@ -2,28 +2,551 @@
...
+        "S2_D2 805",
+        "S2_D2 830",
+        "S1_D2 780",
+        "S1_D2 805",
+        "S1_D2 830",
+        "S1_D1 780",
+        "S1_D1 805",
+        "S1_D1 830",
+    ]
+    assert_array_equal(raw.ch_names[:9], ch_names)
```
larsoner: These should both be lists (I think?) and pytest overwrites `assert` for native types to do nice stuff (nicer than `assert_array_equal` for `list of str` at least) so
```suggestion
    assert raw.ch_names[:9] == ch_names
```

mne_nirs/preprocessing/tests/test_quality.py
```diff
@@ -2,28 +2,551 @@
...
+
+@pytest.fixture(name="fnirs_datasets")
+def fixture_fnirs_datasets(
+    fnirs_motor_data, fnirs_labnirs_3wl_data
+) -> list[mne.io.BaseRaw]:
+    return [fnirs_motor_data, fnirs_labnirs_3wl_data]
+
+
+@mne.datasets.testing.requires_testing_data
+def test_peak_power_runs(fnirs_datasets: list[mne.io.BaseRaw]) -> None:
```
larsoner: If you want a bit more pytest magic... you can avoid repeated use of the `requires_testing_data` decorator with something like this
```suggestion
def fnirs_datasets(
    fnirs_motor_data, fnirs_labnirs_3wl_data
) -> list[mne.io.BaseRaw]:
    if not datasets.has_dataset("testing"):
         pytest.skip("Requires testing dataset")
    return [fnirs_motor_data, fnirs_labnirs_3wl_data]


def test_peak_power_runs(fnirs_datasets: list[mne.io.BaseRaw]) -> None:
```

mne_nirs/preprocessing/_peak_power.py
```diff
@@ -54,6 +54,14 @@ def peak_power(
         List of the start and end times of each window used to compute the
         peak spectral power.
 
+    Notes
+    -----
+    This implementation of peak power differs from Pollonini's original[1]_[2]_,
```
larsoner: This doesn't [render properly](https://output.circle-artifacts.com/output/job/5dabe4b8-f6f5-4d36-ac18-a8532ff3beb7/artifacts/0/html/generated/mne_nirs.preprocessing.peak_power.html)

Image

I'm about 70% sure spacing them out will fix it
```suggestion
    This implementation of peak power differs from Pollonini's original [1]_ [2]_,
```

mne_nirs/preprocessing/_scalp_coupling_segmented.py
```diff
@@ -46,12 +46,20 @@ def scalp_coupling_index_windowed(
...
+        Array of SCI values.
     times : list
-        List of the start and end times of each window used to compute the
-        peak spectral power.
+        List of the start and end times of each window used to compute SCI.
+
+    Notes
+    -----
+    This implementation of scalp coupling index differs from Pollonini's
+    original[1]_[2]_, in that the original calculates scalp coupling index on raw
```
larsoner:
```suggestion
    original [1]_ [2]_, in that the original calculates scalp coupling index on raw
```

larsoner (commented):
> By the way, I'd appreciate feedback about the changes to how channel ordering is handled, whether the previous behaviour is actually a bug or I misunderstood how the functions worked.

I looked through the changes but didn't immediately see what you mean 😅 Could you make a comment inline? But in general I like the change to the channel group indexing rather than using `picks[::2]` type stuff

> I want to see weather you would need to make any changes to "make html" on the examples there that use _peak_power.py

@zEdS15B3GCwq can you look at the examples on CircleCI and compare to https://mne.tools/mne-nirs/stable/index.html ? I would expect them not to change since the changes here should just fix support for multi-wavelength files

mne_nirs/preprocessing/tests/test_quality.py
zEdS15B3GCwq: OK

mne_nirs/preprocessing/tests/test_quality.py
zEdS15B3GCwq: Understood. I took that from `test_snirf.py` and similar code in MNE-Python.

mne_nirs/preprocessing/tests/test_quality.py
zEdS15B3GCwq: Indeed.

mne_nirs/preprocessing/tests/test_quality.py
zEdS15B3GCwq: OK. I didn't feel confident enough to not use it the same way I saw in the few test files I checked.

mne_nirs/preprocessing/_peak_power.py
zEdS15B3GCwq: AI buddy agrees with you. Let's give it a try.

zEdS15B3GCwq:
@larsoner, regarding the `@pytest.fixture` decorators, I tested the code without them, as suggested, but I kept getting errors like this:

"""
_________________________________________________________ ERROR at setup of test_peak_power_known_values_multi_wavelength _________________________________________________________
file C:\Code\mne-nirs\mne_nirs\preprocessing\tests\test_quality.py, line 379
  @mne.datasets.testing.requires_testing_data
  def test_peak_power_known_values_multi_wavelength(
E       fixture 'fnirs_labnirs_3wl_data' not found
>       available fixtures: anyio_backend, anyio_backend_name, anyio_backend_options, cache, capfd, capfdbinary, caplog, capsys, capsysbinary, capteesys, close_all, cov, doctest_namespace, free_tcp_port, free_tcp_port_factory, free_udp_port, free_udp_port_factory, matplotlib_config, monkeypatch, no_cover, options_3d, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, requires_pyvista, subtests, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory
>       use 'pytest --fixtures [testpath]' for help on them.
"""

It seems that pytest did not implicitly treat them as fixtures after all, so in the end I reverted those changes to avoid getting errors.

Another reason I use the pattern of giving the function a `fixture_` prefix and name the fixture in the decorator, is to keep pylint (and perhaps other linters) from complaining about redefined external variables, as they don't understand how pytest uses fixtures.

larsoner:
Oof sorry I didn't meant to suggest removing the decorator. You still need that. What I meant to say is that you do not need to prefix the name with `fixture_`. Sorry about that!

zEdS15B3GCwq:
As for the FIX changes you asked about, it's about the original code not handling properly the case when channel names in the `raw` dataset are not sorted.

1. `plot_3d_montage` uses the [::2] stride, which would only work if the channels are sorted as groups (for 2 wavalengths: ABABAB where A is one wavelength, B is the other; but A and B can be reversed such as ABBA). If, for example, channels are listed as AAAAA first then BBBB, then it can easily miss some channels. E.g. if the channels are S1_D1 780, S1_D2 780, S1_D1 860, S1_D2 860, then it would only pick S1_D1 channels and miss S1_D2 entirely.

2. peak power and sci use `picks` to target the correct channels, and these picks are sorted alphabetically (by `_check_channels_ordered` in mne-python's `nirs.py`), so the problem is different here. The quality measures are calculated correctly as they follow the `picks` indices, however, BAD marks do not use `picks`, just the pure `ii` index. This means that BAD annotations are assigned as 0, 1, 2, 3... order regardless of what picks are used:

```python
        for ii in range(0, len(picks), 2):
            [...]
            c1 = filtered_data[picks[ii]][start_sample:end_sample]  # 1. correct, based on picks
            [...]
           scores[ii, window] = max(pxx)  # 2a. correct, with 2b sorting it later
            [...]
            if (threshold is not None) & (max(pxx) < threshold):
                raw.annotations.append(
                    t_start,
                    time_window,
                    "BAD_PeakPower",
                    ch_names=[raw.ch_names[ii : ii + 2]],  # 3. INCORRECT, doesn't respect picks
                )   
    scores = scores[np.argsort(picks)]  # 2b. this is why the scores assignmemt is correct
```

larsoner:
> it's about the original code not handling properly the case when channel names in the raw dataset are not sorted.

Yes I think the original code assumed some things about the channel ordering that were not necessarily true. So getting the paired indices once and using them consistently is a good fix :+1: 

Let me know if you want me to push a commit to fix the decorator stuff that I made you break :sweat_smile:

zEdS15B3GCwq:
> Yes I think the original code assumed some things about the channel ordering that were not necessarily true.

That's what I needed to know.

> Let me know if you want me to push a commit to fix the decorator stuff that I made you break 😅

It was all a big misunderstanding. ¯\\_(ツ)_/¯

Nevertheless, as I mentioned earlier, if I don't give the fixture function a different name (i.e. prefixed with `fixture_`) than the actual fixture name used in function parameters, then linters (pylint for sure) see it as a problem. See https://pylint.pycqa.org/en/latest/user_guide/messages/warning/redefined-outer-name.html. I reverted those changes to my original version and I'd prefer to keep it that way if possible.

mne_nirs/preprocessing/tests/test_quality.py
```diff
@@ -2,28 +2,548 @@
 #
 # License: BSD (3-clause)
 
-import os
+from pathlib import Path
 
-import mne
+import mne  # type: ignore
```
zEdS15B3GCwq, line 7: I added the comment to silence type checkers, can remove if necessary.

mne_nirs/preprocessing/tests/test_quality.py
```diff
@@ -2,28 +2,548 @@
...
+    return mne.preprocessing.nirs.optical_density(raw)
 
-    raw, scores, times = peak_power(raw)
-    assert len(scores) == len(raw.ch_names)
 
+@pytest.fixture(name="fnirs_labnirs_3wl_data")
+def fixture_fnirs_labnirs_3wl_data() -> mne.io.BaseRaw:
+    """Read and return 3-wavelength testing data."""
+    if not mne.datasets.has_dataset("testing"):
+        pytest.skip("Requires testing dataset")
```
zEdS15B3GCwq: Only the 3-wavelength dataset is from the testing data, the other doesn't need this check.

zEdS15B3GCwq:
Some tests were completely skipped when the testing dataset wasn't available, whereas they could've still used the motor dataset. I pushed a change to fix that, so now only one branch that uses the testing data fails.

By the way, the tests `test_pp_windowed_annotations_target_correct_channels` and `test_sci_windowed_annotations_target_correct_channels` are meant to detect when the bad marks are assigned to the wrong channels. If you run these with the earlier code, both pp and sci will fail.

larsoner:
Ready for merge from your end @zEdS15B3GCwq ? Still see a `[WIP]` in the title but I'm assuming that's outdated now?

zEdS15B3GCwq:
> Ready for merge from your end @zEdS15B3GCwq ? Still see a `[WIP]` in the title but I'm assuming that's outdated now?

Yes, ready if you also think it is.

mne_nirs/preprocessing/tests/test_quality.py
larsoner: I don't think we've thought much about typing here. I don't think we need it over in MNE-BIDS-Pipeline where we have been (much) better about typing. But we can have it for now and fix / change it later if needed

larsoner (approved):

larsoner:
Marking for merge-when-green and starting CIs, thanks in advance @zEdS15B3GCwq !

PR #714, opened 2026-09-03
ENH: Add Homer3 motion artifact detection and correction

leonardozaggia:
#### Reference issue (if any)

None (new feature). Follows the discussion in mne-tools/mne-python#13693, where it was agreed that new fNIRS algorithms should live in mne-nirs. Supersedes mne-tools/mne-python#13692 and mne-tools/mne-python#13693.

Adds Homer3-based motion artifact detection and correction to `mne_nirs.preprocessing`:

- `detect_motion_artifacts` / `detect_motion_artifacts_by_channel` (aliases `motion_artifact` / `motion_artifact_by_channel`): flag samples where the signal changes by more than `stdev_thresh` times the standard deviation of its derivative, or by more than `amp_thresh`, within a window of `t_motion` seconds, padded by `t_mask` seconds. The per-channel version evaluates both wavelengths of a source-detector pair together.
- `motion_correct_spline` (alias `spline`): detrends each motion segment with a smoothing spline and baseline-shifts consecutive segments so they connect smoothly. Accepts a per-channel mask; when None it is built from `BAD` annotations or by running the detection with default parameters.
- `motion_correct_wavelet` (alias `wavelet`): pads each channel to a power of 2, applies Homer3's translation-invariant wavelet transform, zeroes detail coefficients that are IQR-based outliers and reconstructs the signal. Specialises in spike removal.
- `motion_detect_and_correct_wavelet` (alias `auto_wavelet`): detection followed by wavelet correction, with the detected segments added as `BAD_motion` annotations.

Based on Homer3 v1.80.2 `hmrR_MotionArtifact`, `hmrR_MotionArtifactByChannel`, `hmrR_MotionCorrectSpline` and `hmrR_MotionCorrectWavelet` (Huppert et al. 2009, Scholkmann et al. 2010, Molavi & Dumont 2012). Adds PyWavelets to the testing and docs requirements; it is imported lazily, so it is not a hard runtime dependency.

The ports reproduce Homer3's output. `test_homer3_equivalence.py` runs the functions on two recordings from the MNE testing dataset and compares them with results computed by Homer3 in MATLAB (tests/data/homer3_motion_correction.npz, 253 kB; the MATLAB driver and the script that build the archive are in this gist: https://gist.github.com/leonardozaggia/9d7b02c0fc854dd7ba0df4fbaae69a9c).
Masks are identical, the spline agrees to 1e-12 (1e-5 on one file whose stored time vector MNE regularises) and the wavelet to 1e-11. The same holds on the full 50-minute fNIRS motor recording.

#### Additional information

- Operates on optical density or hemoglobin data.
- `smoothing` is Homer3's `p` with `csaps` semantics (1 = interpolation, 0 = straight line), implemented with `scipy.interpolate.make_smoothing_spline`; default 0.99 as recommended in the literature. This needs scipy >= 1.10, so the floor in `pyproject.toml` is raised.
- `mask` is shape (n_picks, n_times); True = clean, False = artifact. The Homer3 names (`tInc`, `tIncCh`, `p`) are noted in the docstrings.
- `iqr` controls the outlier threshold; set to -1 to skip the correction entirely.
- New sections in `plot_21_artifacts.py` show detection, spline and wavelet correction on the injected spike and baseline shift.
- The PCA motion correction from mne-tools/mne-python#13691 is independent work and will come in a separate PR.


mne_nirs/preprocessing/motion_removal/_motion_artifact.py
```diff
@@ -0,0 +1,305 @@
...
+    raw : instance of Raw
+        Raw object whose annotations are inspected.
+    n_picks : int
+        Number of fNIRS channels.
+    n_times : int
+        Number of time samples.
+
+    Returns
+    -------
+    mask : ndarray of bool, shape (n_picks, n_times)
```
larsoner, line 59: You'll need to add `n_picks` and `n_times` to `doc/conf.py:numpydoc_xref_ignore` to make CircleCI happy

larsoner (commented):
Just one tiny comment to hopefully get CIs happy, I should be able to review more completely soon!

mne_nirs/preprocessing/tests/data/make_homer3_reference.py
```diff
@@ -0,0 +1,116 @@
+# Authors: The MNE-Python contributors.
```
larsoner: All this stuff won't be run by us... can you put it in a public gist on `gist.github.com` and link to it here? That's what we've done in MNE-Python when we make "equivalence with package X" parts of PRs

mne_nirs/preprocessing/motion_removal/_motion_artifact.py
leonardozaggia: Thanks, added both to `numpydoc_xref_ignore`

mne_nirs/preprocessing/tests/data/make_homer3_reference.py
leonardozaggia: Moved both scripts out of the package into a public gist: https://gist.github.com/leonardozaggia/9d7b02c0fc854dd7ba0df4fbaae69a9c. Only the 253 kB reference archive stays in the repo, and the test module docstring now points readers to the gist via this PR.

larsoner:
@leonardozaggia did you use an LLM to help with any of this? If so, can you have it read https://github.com/mne-tools/mne-python/blob/main/AGENTS.md and rework the PR appropriately? +1700 lines seems like it could potentially be LLM code bloat... but in any case, it would be really nice if we could find a way to make the implementation and tests much more compact, especially if we can make use of SciPy / NumPy functions instead of constructing bespoke helpers, etc.

PR #719, opened 2026-09-21
BUG: Fix windowed quality metrics on Raws with non-fNIRS channels

### LINKED ISSUE #718: BUG: Windowed quality metrics fail on Raws containing non-fNIRS channels
## Description of the problem

In `scalp_coupling_index_windowed()` and `peak_power()`, the score array is
allocated with one row per fNIRS channel but addressed by raw channel index:

```python
# mne_nirs/preprocessing/_scalp_coupling_segmented.py, same in _peak_power.py
picks = _validate_nirs_info(raw.info, fnirs="od", which="Scalp coupling index")
sci = np.zeros((len(picks), n_windows))     # one row per fNIRS channel
...
ch_group = picks[gg : gg + n_wavelengths]   # raw ch indices
sci[ch_group, window] = c                   # raw index into a pick-sized array
```

--> Index mismatch if data includes non-fnirs channels.

## Steps to reproduce

```python
import mne
import numpy as np

from mne_nirs.preprocessing import scalp_coupling_index_windowed

names = ["Trig", "S1_D1 760", "S1_D1 850", "S10_D10 760", "S10_D10 850"]
info = mne.create_info(names, 10.0, ["stim"] + ["fnirs_od"] * 4)
for ch in info["chs"][1:]:
    pair, wavelength = ch["ch_name"].rsplit(" ", 1)
    idx = 0 if pair == "S1_D1" else 1
    ch["loc"][:3] = ch["loc"][3:6] = (0.01 * idx, 0, 0)
    ch["loc"][6:9] = (0.01 * idx + 0.03, 0, 0)
    ch["loc"][9] = float(wavelength)

rng = np.random.RandomState(0)
raw = mne.io.RawArray(rng.randn(len(names), 300) * 1e-3, info, verbose=False)

scalp_coupling_index_windowed(raw, time_window=10, threshold=0.5)
```
## Expected results

Metrics computed over the fNIRS channels, ignoring other channel types.

## Actual results

```text
IndexError: index 4 is out of bounds for axis 0 with size 4
```

`peak_power()` fails the same way.

## Possible fixes

The `scores` (sci or pp) are currently in a `(n_nirs_channels, n_windows)` array. 
Would it be better to:
1. Keep this scores array format and fix the local index issue. This also requires fixing downstream functions that assume that the scores array channels match raw channels, like `plot_timechannel_quality_metric`. Also, the order of the `scores` channels does not necessarily match the order of `mne.pick_types(info, fnirs=True)` so pairing the scores to the raw channels is bug prone.
2. Change `scores` array shape to one row per one Raw channel and have NaN for non-fnirs channels. This may be a bit ugly but would be much easier to use as it does not require the user to manually pair the scores to Raw channels. Also would not need to touch downstream functions.

## Additional information

```text
Platform             macOS-26.2-arm64-arm-64bit
Python               3.12.13 (main, Mar  3 2026, 15:35:03) [Clang 21.1.4 ]
Executable           /Users/kallemakela/code/fnirs/external-packages/mne-nirs/.venv/bin/python
CPU                  Apple M1 Pro (8 cores)
Memory               32.0 GiB

Core
├☑ mne               1.13.2 (latest release)
├☑ numpy             2.5.3 (Accelerate, threads not introspectable)
├☑ scipy             1.18.1
└☑ matplotlib        3.11.2 (backend=macosx)

Numerical (optional)
├☑ scikit-learn      1.9.1
├☑ threadpoolctl     3.7.0
├☑ nibabel           5.4.2
├☑ nilearn           0.14.1
├☑ pandas            3.0.6
├☑ h5io              0.2.5
├☑ h5py              3.16.0
└☐ unavailable       numba, dipy, openmeeg, python-picard, jamica, cupy

Visualization (optional)
└☐ unavailable       pyvista, pyvistaqt, vtk, qtpy, ipympl, pyqtgraph, mne-qt-browser, ipywidgets, trame, trame_client, trame_server, trame_pyvista, trame_vtk, trame_vuetify

Ecosystem (optional)
├☑ mne-nirs          0.8.0.dev71+g4998adc9d
└☐ unavailable       mne-bids, mne-features, mne-connectivity, mne-icalabel, mne-bids-pipeline, autoreject, neo, eeglabio, edfio, curryreader, mffpy, pybv, pymef, antio, defusedxml
```

kalomak:
Closes #718.

`scalp_coupling_index_windowed()` and `peak_power()` sized the score array by the
number of fNIRS channels but wrote into it with raw channel indices, so any Raw
with a non-fNIRS channel ahead of the fNIRS ones raised `IndexError`.

Both now index the score array pick-relative inside the loop and restore the
original channel order on the way out, the same way
`mne.preprocessing.nirs.scalp_coupling_index` does (not ideal).


larsoner (approved):
Looks good, thanks @kalomak !

